PR #1, opened 2016-01-13
Add autoprefixer

msecret:
Not fully tested yet


afeld:
Not sure if you want it tested yet, but looks good to me!

PR #2, opened 2016-01-13
Fix var overrides

msecret:
```
Temporary fix USWDS default variables

This temporarily changes the USWDS dependency to my own branch so
that variables can be override. More information on the problem here:
https://github.com/18F/web-design-standards/issues/952

Can be reverted back when this is complete:
https://github.com/18F/web-design-standards/pull/953.
```

### DIFF
--- a/css/cg_style.css
+++ b/css/cg_style.css
@@ -176,7 +176,6 @@ figure {
  * Address differences between Firefox and other browsers.
  */
 hr {
-  -moz-box-sizing: content-box;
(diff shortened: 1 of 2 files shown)


afeld:
:shipit: 

Not merging because I wasn't sure if you meant for this pull request to be done against `master`...?

msecret:
Yea @afeld you're correct, I wanted it done against staging, will redo

PR #7, opened 2016-01-21
Sidenav style

jcscottiii:
Used in https://github.com/18F/cg-docs/pull/183


jcscottiii:
Used in https://github.com/18F/cg-docs/pull/183

PR #13, opened 2016-01-25
Mobile and all

msecret:
Combination of billboard, override, ups, footer, adds
mobile styles to ensure things look ok on mobile to finish up
the whole landing homepage.

Probably should merge this and forget the rest of the PRs.

### DIFF
--- a/src/css/base/form.scss
+++ b/src/css/base/form.scss
@@ -14,7 +14,7 @@ label {
 
 input[type="email"] {
   font-size: 1rem;
-  padding: .7rem 0.7em;
+  padding: 0.7rem 0.7rem;
 }
 
 form button[type="submit"], 
--- a/src/css/base/form.scss
+++ b/src/css/base/form.scss
@@ -1,0 +1,23 @@
+
+form {
+  max-width: none;
+}
+
+legend {
+  font-size: 1.2rem;
+  margin-bottom: .5rem;
+}
+
+label {
+  margin-top: 0;
+}
+
+input[type="email"] {
+  font-size: 1rem;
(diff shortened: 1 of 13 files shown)


src/css/base/typography.scss
```diff
@@ -7,6 +7,10 @@ h1, h2, h3, h4, h5, h6 {
   & + p {
     margin-top: 0.5rem;
   }
+
+  & + .push {
```
jcscottiii: I see only the landing page is using this for contact. 

hmm. usually when i see push / pull classes I see it as `push-X` x being a number of some sorts. just want to know how much to expect the push in the name.

jcscottiii:
maybe `four_up` -> `four-cols` or `four-columns` and 
 `four_up-part` -> `four_cols-col` or `four_columns-col`?
and similar scheme for three_up too.

src/css/base/form.scss
```diff
@@ -0,0 +1,23 @@
...
+  margin-bottom: .5rem;
+}
+
+label {
+  margin-top: 0;
+}
+
+input[type="email"] {
+  font-size: 1rem;
+  padding: .7rem 0.7em;
```
jcscottiii: use rem for the second one? and specify all four dimensions or explicitly only do `padding-top` and `padding-right`

jcscottiii:
@msecret just a few suggestions above. otherwise looks good!

src/css/base/typography.scss
msecret: Yea I want a different name to it because it's different then the standard push. It's a class that signifies that a div should be treated as a block of text.

PR #17, opened 2016-01-26
Feature hex icons

msecret:
Look at commit messages for more information.


src/css/components/icon.scss
```diff
@@ -0,0 +1,15 @@
...
+  height: 3.125rem;
+  width: 3.9rem;
+}
+
+.icon {
+  max-height: 3.1rem;
+}
+
+.icon-alt {
+  .cls-1 {
```
jcscottiii, line 12: I see this is everywhere in the .svg. Did they automatically give you that name? If not, what about `.cls-stroke`?

msecret, line 12: Yea it was automatically generated by illustrator.

I almost thought it would be easier to use that class for now rather then editing every single svg.

I think a better system eventually needs to be put in place for svg, but this will do for launch

jcscottiii, line 12: Make an issue about it @msecret. Sounds good. Will merge.

PR #23, opened 2016-01-29
Separators under headings with hexagon

msecret:
These were built so they could conform to many different widths
and the middle piece could be switched out. With this design, the
width of the middle piece can be changed and the lines will conform.
The hexagon class uses ems for this reason, so you can set the font
size to control it's size. This had to be done because borders can't
be set with percentages so ems are used instead. And then the size
of the hexagon is set with the parent font-size.

It can also change colors from primary to alt or any other.

### DIFF
--- a/src/css/components/hexagon.scss
+++ b/src/css/components/hexagon.scss
@@ -1,0 +1,47 @@
+
+$primary-color: $color-primary-alt;
+$alt-color: $color-secondary;
+
+.hexagon {
+	background: $primary-color;
+  height: .65em;
+	position: relative;
+  width: 1em;
+}
+
+
+.hexagon:before {
+	border-left: .5em solid transparent;
+	border-right: .5em solid transparent;
+	border-bottom: .25em solid $primary-color;
+	content: "";
+	position: absolute;
+  top: -.25em;
+	left: 0;
+	width: 0;
+	height: 0;
+}
+
+.hexagon:after {
+	border-left: .5em solid transparent;
+	border-right: .5em solid transparent;
+	border-top: .25em solid $primary-color;
+	content: "";
+	position: absolute;
+	bottom: -.25em;
+	left: 0;
+	width: 0;
+	height: 0;
+}
+	
+.hexagon-alt {
+	background: $alt-color;
+
+  &:before {
+    border-bottom-color: $alt-color;
+  }
+
+  &:after {
+    border-top-color: $alt-color;
+  }
+}
--- a/src/css/components/hexagon.scss
+++ b/src/css/components/hexagon.scss
@@ -3,39 +3,35 @@ $primary-color: $color-primary-alt;
 $alt-color: $color-secondary;
 
 .hexagon {
-	background: $primary-color;
+  background: $primary-color;
   height: .65em;
-	position: relative;
+  position: relative;
   width: 1em;
 }
 
+.hexagon:before,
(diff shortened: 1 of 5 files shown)


src/css/components/hexagon.scss
```diff
@@ -0,0 +1,47 @@
+
+$primary-color: $color-primary-alt;
+$alt-color: $color-secondary;
+
+.hexagon {
+	background: $primary-color;
+  height: .65em;
```
jcscottiii, line 7: uneven spacing between lines

src/css/components/hexagon.scss
```diff
@@ -0,0 +1,47 @@
...
+}
+
+
+.hexagon:before {
+	border-left: .5em solid transparent;
+	border-right: .5em solid transparent;
+	border-bottom: .25em solid $primary-color;
+	content: "";
+	position: absolute;
+  top: -.25em;
```
jcscottiii, line 25: same here

jcscottiii, line 25: well this line has the right spacing, but not the others around it

src/css/components/hexagon.scss
```diff
@@ -0,0 +1,47 @@
...
+	content: "";
+	position: absolute;
+	bottom: -.25em;
+	left: 0;
+	width: 0;
+	height: 0;
+}
+	
+.hexagon-alt {
+	background: $alt-color;
```
jcscottiii: spacing

src/css/components/hexagon.scss
```diff
@@ -0,0 +1,47 @@
+
+$primary-color: $color-primary-alt;
+$alt-color: $color-secondary;
+
+.hexagon {
+	background: $primary-color;
+  height: .65em;
+	position: relative;
+  width: 1em;
```
jcscottiii, line 9: em -> rem?

src/css/components/hexagon.scss
```diff
@@ -0,0 +1,47 @@
...
+}
+
+
+.hexagon:before {
+	border-left: .5em solid transparent;
```
jcscottiii: em -> rem?

src/css/components/hexagon.scss
```diff
@@ -0,0 +1,47 @@
...
+	content: "";
+	position: absolute;
+  top: -.25em;
+	left: 0;
+	width: 0;
+	height: 0;
+}
+
+.hexagon:after {
+	border-left: .5em solid transparent;
```
jcscottiii: em -> rem

src/css/components/separator.scss
```diff
@@ -0,0 +1,47 @@
...
+    background: linear-gradient(to top, $alt-color 5%, 
+      rgba($alt-color, 0.1) 90%);
+  }
+}
+
+.separator-center_piece {
+  flex-grow: 0;
+  flex-shrink: 1;
+  font-size: $center_piece_size;
+  margin-left: 1em;
```
jcscottiii: em -> rem?

src/css/components/hexagon.scss
msecret, line 9: These are actually all supposed to be `em` because you can set teh font-size of the parent to control the size of the hexagon. This was done rather then using percentage units for everything because border sizes can't be defined with percentages but can be defined with `em`s.

PR #24, opened 2016-01-29
Typography

msecret:


src/css/base/typography.scss
```diff
@@ -53,12 +61,10 @@ h5 {
...
 
-body {
-  font-weight: 300;
-  line-height: 1.7333333;
-}
-
 p {
+  font-family: $font-serif;
+  font-size: 1.05rem;
+  line-height: 1.75;
```
jcscottiii, line 67: unit?

src/css/base/typography.scss
```diff
@@ -15,9 +21,11 @@ h1, h2, h3, h4, h5, h6 {
 
 h1 {
   font-size: 2rem;
+  line-height: 1.19;
 
   @include media($medium-screen) {
     font-size: 2.75rem;
+    line-height: 1.18;
```
jcscottiii, line 28: unit?

src/css/base/typography.scss
```diff
@@ -15,9 +21,11 @@ h1, h2, h3, h4, h5, h6 {
 
 h1 {
   font-size: 2rem;
+  line-height: 1.19;
```
jcscottiii, line 24: unit?

src/css/base/typography.scss
```diff
@@ -1,4 +1,10 @@
 
+body {
+  color: $color-base;
+  font-weight: 300;
+  line-height: 1.7333333;
```
jcscottiii, line 5: unit?

src/css/base/typography.scss
msecret, line 67: When line-height doesn't have a unit, it makes it relative to the font-size of the current element, which is what I wanted.

PR #25, opened 2016-02-02
Background section chevrons

msecret:
These are the v shaped design of the sections.

This was done by putting svg shapes on the front and back
of the sections and specifying classes for what type
of shape at what point.

This was done with svg because doing CSS shapes at 100% width
like this seemed difficult. This was all done in the section
classes because these shapes are really part of the sections,
or the sections are these shapes.

TODO:
- Investigate way to keep svg to 100% width but also maintain
  aspect ration to ensure v shape doesn't change.

### DIFF
--- a/src/css/components/section.scss
+++ b/src/css/components/section.scss
@@ -1,13 +1,12 @@
 
 .section {
-  overflow: hidden;
-  padding-bottom: 4.5rem;
-  padding-top: 4.5rem;
+  padding-bottom: 12.5rem;
+  padding-top: 2.75rem;
 }
 
 .section-middle {
-  padding-bottom: 4.5rem;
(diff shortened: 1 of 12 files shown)


src/css/components/section.scss
```diff
@@ -74,11 +87,78 @@
...
+  }
+
+  &:not(.section-first) {
+    &:before {
+      background-image: url("../img/chevron-top-inverse.svg");
+      top: -5rem;
+    }
+  }
+
+  /*
```
jcscottiii: Remove or use commented-out style

jcscottiii:
Will you be addressing the TODO in this PR btw? If not, make an issue

msecret:
I figured out it won't be addressed at all.

PR #36, opened 2016-02-17
Changes names of assets and library

msecret:
Changes npm package to be cloudgov-style. Also renames a bunch of the
assets to be consistent.

This is a breaking chnage, sites using cg-style have to be updated. The
gem also has to be updated with this.


Gemfile.lock
```diff
@@ -1,5 +1,7 @@
-PATH
-  remote: ./gem
+GIT
+  remote: git://github.com/18F/cloudgov-style.git
```
jcscottiii: when you rename this repo, i'll be fine with this PR.

Gemfile
```diff
@@ -7,5 +7,5 @@ gem 'redcarpet'
 gem 'sass'
 
 group :jekyll_plugins do
-  gem 'cloudgov-style'
+  gem 'cloudgov-style', :github => '18F/cloudgov-style', :branch => 'master'
```
jcscottiii: fix repo name

Gemfile.lock
jcscottiii: after talking, we aren't changing this, needs to be changed back

PR #39, opened 2016-02-17
Merge changes into live styleguide

msecret:
This might need a better process in the future.


msecret:
screen shot 2016-02-23 at 3 05 01 pm

jeremiak:
This looks good to me and we want to keep our public documentation up to date

PR #47, opened 2016-04-06
Webpack var for assets

msecret:
In order to use webpack with cg-style, the paths to images and fonts in `url`s have to be relative to the base scss file that's being imported. This PR sets up variables to override the paths of `url`s so a consumer using webpack can set them to the correct path (current `./img` and `./font` in this project).

See commit messages for details


src/css/components/hex-icon.scss
```diff
@@ -14,9 +14,9 @@
 }
 
 .hex_icon-primary {
-  background-image: url("../img/icon-hex-primary.svg");
+  background-image: url("#{$static-img-path}/img/icon-hex-primary.svg");
```
msecret: Remove extra `/img`.

src/css/override_vars.scss
```diff
@@ -1,4 +1,8 @@
 
+// Path to override for web design standards
+$static-img-path: '../img' !default;
+$static-font-path: '../font' !default;
```
msecret: Shouldn't the standards font path be `fonts`?

src/css/override_vars.scss
```diff
@@ -1,4 +1,8 @@
 
+// Path to override for web design standards
+$static-img-path: '../img' !default;
+$static-font-path: '../fonts' !default;
```
jcscottiii, line 4: should this be singular `font` like the rest of the references?

msecret, line 4: Because it's the overrides to the standards, it's `fonts` because the standards is `fonts`.

It's confusing because I need to be able to override both the standards and cg-style paths.

PR #49, opened 2016-04-15
Initial build of components

msecret:
Building the individual sass components as CSS components so it's easier
to import them individually into sites with webpack.

This outputs:

```
dist/
- components/
- - billboard.css
- - ...
- base.css
- cloudgov-style.css
```

TODO
- Want to use include paths to limit dipping into node_modules


jeremiak:
Great, looks good to me. Merging because its done and I need #50 for 18f/cg-deck#173

PR #50, opened 2016-04-19
Ms correct paths in base

msecret:
Check 0c5a4aa for overview of code

This branch is inheriting from #49, that should be merged first. 

https://github.com/18F/cg-deck/issues/173 is dependent on this.


jeremiak:
Looks good, runs good!

PR #51, opened 2016-04-19
Ms improve tables

msecret:
Check 159eccc for review of what was done.

Tested on table in cloudgov docs and cloudgov deck.

Build on branch #50 that should be merged first.

Tracked by #54

PR #55, opened 2016-04-21
Add documentation for table

jeremiak:
In reviewing #51 I noticed that there isn't documentation for the base styles, though it would definitely be useful for cases like this. I've gone ahead and done a super basic breakup just to show the direction I was thinking and get feedback. If we like this, I can finish it up before this PR is merged.

This PR so far:
- Add documentation for table base style and creates a "base" collection
- Changes the navbar to only have base and component links. Each goes to a collection index page that has a sidenav element with all of the collection items.

Screenshot: 

Before this is merged we should:
- [x] Agree that we want to go in this direction for documentation
- [ ] Fix nav bar stylings so that two elements aren't so far apart
- [ ] Add sidenav to each item within a collection for easier navigation
- [ ] Have the index page of each collection show all the elements in one long list

@msecret what do you think?

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -30,6 +30,9 @@ sass:
 #  style: :compressed
 
 collections:
+  base:
+    output: true
+    permalink: /:path/
   components:
     output: true
     permalink: /:path/
--- a/documentation/_base/table.md
+++ b/documentation/_base/table.md
@@ -1,0 +1,34 @@
+---
+layout: component
+type: component
+title: Table
+---
+
+<table>
+  <thead>
+    <tr>
+      <th>Column 1</th>
+      <th>Column 2</th>
+      <th>Column 3</th>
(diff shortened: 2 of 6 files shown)


msecret:
On the first two points:
- I definitely like this direction
- I'm not sure I care about the nav bar styling right now, this site is for testing right now so doesn't necessarily need to look nice

The next two sound cool but not sure how necessary.

jeremiak:
Nice, I mostly agree @msecret. I think it will help us realize problems with the styles in general though (the top navbar, for example, should handle X number of items gracefully and not be too far apart). That might be a bit of yack shaving right now because we _do_ have more than just two items on the production versions of this.

So in that case, want to give this a review/merge? We can open other issues later to deal with any other styling issues we care about.

PR #57, opened 2016-04-22
Add font awesome as a depedency

jeremiak:
In the process of documenting how the sidenav element works I noticed we didn’t have font-awesome included which resulted in the glyphs not being displayed properly. Added the library as a dependency and included some build scripts that are run automatically on `npm run build`

The `copy-fontawesome-css` script is weird looking because I alter the paths required in the CSS file to match our current file structure. We currently use `/font` so that is where the browser should look for font awesome fonts.

I raised some related concerns in 18F/cg-style-gem#5, which could change some parts of this implementation

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "The global style of cloud.gov",
   "main": "src/js/main.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


jeremiak:
I'm closing this PR because it isn't the route we want to go. Since we're pulling in font-awesome just for some carets for the sidenav, we are just going to style the carets with pure CSS to reduce the footprint of using `cg-style`.

PR #58, opened 2016-04-22
Document a more complex sidenav example

jeremiak:
The documentation now includes nested menus for easier reference of how the sidenav element is styled. To accomplish this I needed:
- To add example code. I used a modified version of the sidenav on [docs.cloud.gov](https://docs.cloud.gov)
- To add `font-awesome` as a dependency and build it into the site
- To add `main.js` from the production docs site to give the sidenav the proper behavior.

This should be merged after #60

### DIFF
--- a/documentation/_components/sidenav.md
+++ b/documentation/_components/sidenav.md
@@ -4,39 +4,79 @@ type: component
 title: Sidenav
 ---
 
-<aside id="sidebar" class="sidenav">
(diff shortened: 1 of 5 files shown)


jeremiak:
Now that we've closed #57 in favor of #60, this should not be merged until #60 has been merged and rebased into this branch. CSS classes will need to change as well.

PR #63, opened 2016-04-28
Release 0.4.x - Importable components

msecret:
Ups version number.


jeremiak:
💥  Let's get a release out!

PR #64, opened 2016-05-09
Searchbar

jcscottiii:
This is for a search bar in the nav

Goes with https://github.com/18F/cg-docs/pull/252

image


jcscottiii:
@jeremiak @msecret

jeremiak:
Looks good @jcscottiii! I think we should also add this to the `documentation/_includes/header.html` partial so that it is used on the style guide site.

Also, how do you feel about moving the `search.scss` into `nav.scss`? With the way React components import styles in 18f/cg-deck its a bit easier to only have to import one file per component and since the search bar will always be in the nav it seems like a good fit.

msecret:
I want to wait til we get Victor involved to merge this. We might want to add the functionality to expose the nav after you clikc on a search icon first.

jeremiak:
If we merge this we should also tag a `0.4.1` release. I'm using patch versions as a stand in for minors, sort of. I think we should just publish the next breaking change as `1.0.0` so we can properly use semver but in the meantime we can stick to our current method.

18F/cg-docs#252 relies on this and it would be best to use a public version of `cloudgov-style` instead of pinning to a specific commit here

src/css/components/search.scss
```diff
@@ -0,0 +1,56 @@
+@import "../core.scss";
+
+$search-btn-width:   4.5rem;
+$search-btn-height:  2.0rem;
+
+@mixin search-icon {
+  background-image: url('../img/i-search.svg');
```
msecret, line 7: Can you use the `$static-img-path` variable here. Example in https://github.com/18F/cg-style/blob/master/src/css/base/background.scss#L15

src/css/components/search.scss
```diff
@@ -0,0 +1,56 @@
...
+  background-position: 50%;
+  background-repeat: no-repeat;
+}
+
+.search-bar {
+  @include clearfix;
+  max-width: none;
+  position: relative;
+  $width: $search-btn-width;
```
msecret, line 16: Can you move this variable to to the top of the selector, line 13.

src/css/components/search.scss
```diff
@@ -0,0 +1,56 @@
...
+@mixin search-icon {
+  background-image: url('../img/i-search.svg');
+  background-position: 50%;
+  background-repeat: no-repeat;
+}
+
+.search-bar {
+  @include clearfix;
+  max-width: none;
+  position: relative;
```
msecret, line 15: Selectors should be alphabetically ordered

src/css/components/search.scss
```diff
@@ -0,0 +1,56 @@
...
+  $width: $search-btn-width;
+  margin-left: 0;
+  margin-top: 0.5rem;
+  margin-right: 0;
+  margin-bottom: 0;
+
+  [type="search"],
+  .search-field {
+    -webkit-appearance: none;
```
msecret, line 24: We're using auto-prefixer, so this should just be `appearance`.

msecret:
@thisisdano @vz3 we could use some design advice/help here.

First off, here's the original issue for implementing the search bar into the header: https://github.com/18F/cg-deck/issues/237. In the first comment, you can see how @jcscottiii implemented this. It doesn't exactly follow the issue as the search bar is longer, and is always present rather then only being present when the search icon is clicked.

Based on this, should we:
- (a) continue with this
- (b) take the extra time to implement the search bar to only appear after you click the search icon (about 1 day of work)
- (c) do something completely different from either (a) or (b).

juliasolorzano:
The design team has decided on a slightly different direction for implementation of the search bar. A new comp to work from is located here: https://github.com/18F/cg-docs/pull/252#issuecomment-244135650

I'll be closing this PR and opening a new one that might branch off of this work. 🎉

PR #66, opened 2016-05-10
Document cg header and navigation component

### LINKED ISSUE #65: Navbar usage and code should be in the style guide documentation
The usage of the navbar component isn't documented in the Jekyll site but should be for easier reference.

The documentation should also note that it [requires some Javascript](/18F/cg-docs/pull/252).

jeremiak:
Fixes #65

Add brief documentation page for the header component, which includes the navigation component.

To do:
- [ ] Document usage with search box & Javascript

### DIFF
--- a/documentation/_base/header.md
+++ b/documentation/_base/header.md
@@ -1,0 +1,65 @@
+---
+layout: component
+type: component
+title: Header & Navigation
+---
+
+<header class="header">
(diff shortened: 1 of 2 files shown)


jeremiak:
We don't know what's happening with the design of the search box just yet, so let's not hold this up on account of that.

@msecret this is ready for review

PR #73, opened 2016-05-13
Release steps

msecret:
Draft of release steps.


RELEASE.md
```diff
@@ -0,0 +1,34 @@
+# Release
+Follow the steps outlined in this document to release a new version of `cloudgov-style`.
+
+### Blah blah
```
msecret: Can't decide on headings and how to structure them.

RELEASE.md
```diff
@@ -0,0 +1,34 @@
...
+4. Commit the changes to the package.json file.
+5. Tag the commit with the release version number in the following format: `git tag -a x.x.x -m 'vx.x.x - Main idea of release`.
+6. Push the release branch to github.
+
+### Ruby gem something
+1. On the release branch, build the gem by running `npm run gem`.
+2. In the `./gem` directory create a branch for the release.
+3. Update the version number in `lib/cloudgov-style/version.rb` to the release version number.
+4. Tag the commit with the release version number in the following format: `git tag -a x.x.x -m 'vx.x.x - Main idea of release`.
+5. Commit the changes and push the branch with the tags.
```
msecret: Actually unsure about this.

So when you tag in github, it'll be attached to a certain commit. So if you tag the release commit and push it, then when the master deploy runs it should all work right?

RELEASE.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Steps
+
+### Preparation
+1. Decide on a "main idea" for the release, what the release is being done for. Look at previous releases for examples.
+2. Decide on whether the release should be a patch, minor or major version update, based on semver.
+3. Create a new release on the Github Releases page.
+  1. Title the release with `{version number} - {main idea}`.
+  2. Use the tag version of the release version number.
+  3. Write a list of changes for the release.
```
msecret: Should this be formatted in a special way? WDS does `fixed`, `changed`, `added`.

RELEASE.md
jeremiak: I think we should describe the point of releasing and we're we'll be publishing to. Its kind of a different sort of release process that tagging this also publishes an npm module and a ruby gem, so we should make that explicit.

RELEASE.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Blah blah
+- Releases should follow [semantic versioning](http://semver.org/).
+- The Github tags for the release should just be the number, no `v` before.
+
+## Steps
+
+### Preparation
+1. Decide on a "main idea" for the release, what the release is being done for. Look at previous releases for examples.
+2. Decide on whether the release should be a patch, minor or major version update, based on semver.
```
jeremiak: Are we still following semver before we release `1.0.0`? That is to say, as long as the next release has no breaking changes and has additional features it will be `0.5.0`?

Relatedly, will the next breaking change be published as `1.0.0`?

RELEASE.md
jeremiak: Sure, I'm open to specified formats

RELEASE.md
msecret: Yes to both. Do we need to change the documentation at all?

RELEASE.md
msecret: I added that we publish to npm and ruby gems, had a hard time writing the point of releasing, feel free to take a try

RELEASE.md
msecret: I decided I'd like to keep it general, and not say specific formats.

RELEASE.md
```diff
@@ -0,0 +1,41 @@
+# Release
+Follow the steps outlined in this document to release a new version of `cloudgov-style`.
+
+### Blah blah
+- Releases should follow [semantic versioning](http://semver.org/).
+- The Github tags for the release should just be the number, no `v` before.
+- The project will be pushed to [npm](https://www.npmjs.com/package/cloudgov-style) and [RubyGems](https://rubygems.org/gems/cloudgov-style).
```
jeremiak, line 7: I'm confused about this, there isn't any sort of automatic publishing of packages using a tag or something like that, correct? By what mechanism does one "push" the project to these two registries?

RELEASE.md
```diff
@@ -0,0 +1,41 @@
...
+  2. Use the tag version of the release version number.
+  3. Write a list of changes for the release.
+  4. Create a link titled `Changelog` that uses the Github compare url to compare the current release tag with the last, such as `/compare/0.3.4...0.4.0`.
+
+### Npm something
+1. Ensure the `master` branch is up to date with all code to release.
+2. Checkout a new `release-x.x` branch with the major and minor numbers of the new release version number.
+  - This release branch should have all the changes to be put into the release.
+  - This branch should be cut from `master`.
+3. Update the `package.json` version number to the release version number.
```
jeremiak: But in this case `master` branch won't have the updated release number in `package.json`. So after you do these steps, would you then open a PR from your new release branch into `master`? In this way, `master` will have the updated version number, but the release tag will be on the commit that updates it on the release branch? Or should the release tag be on the commit that merges the new version number back into `master`?

Or we could just rebase the release branch onto master and `push --force`, but that means we can't have `master` branch be a protected branch.

RELEASE.md
jeremiak: This format makes it seem like you can run either publish the node module or the Ruby gem but in reality you should published both at the same time, right? In fact, the version numbers should maybe be updated in both `package.json` and `lib/cloudgov-style/version.rb` in the same commit?

How can we get this automated so there is something like `npm run publish` which would handle all of this?

msecret: They can't be updated in the same commit because they're actually in different repos. We could combine the repos, which might be the best long term solution.

I like the idea of automating this with a command.

RELEASE.md
msecret, line 7: Yes, so the logic is when there's a new tag, when a PR gets merged to master, travis will deploy.

RELEASE.md
msecret: Thats true, I think the workflow is you put up a PR for the release branch which then gets merged into master, so at that point it has the updated version number.

jeremiak:
Nice job with this draft. I think its generally pretty good but we should present the steps in a way that its clear you should always be publishing both a Node module and Ruby gem when there is a version change. Ideally we could wrap up the publish steps in an `npm` script, but having them written out as an integrated set of steps would be more user friendly.

Also, I think there's still an open question on which commit the release tag gets applied to. One option we have is to keep `master` an "unprotected" branch and just rebase release branches into `master` and `git push --force`.

It seems like the another option is to tag the commit in the release branch where the version numbers are bumped in both `package.json` and `lib/cloudgov-style/version.rb` and then open a PR against master. In this case the merge commit into master will not be the release commit.

Or we could just tag the resulting merge commit into `master` from above as the release. 

I personally like rebasing the release commits back into `master` and force pushing since it will keep the history clean and the release commits will be obvious and be reflected in the commit messages but I'm open to any of the processes above.

Thoughts or preferences?

RELEASE.md
jeremiak: No, I don't think this documentation needs to change at all. Since you or I will likely be doing that release we can just make sure to bump the next breaking change to `1.0.0`

RELEASE.md
msecret: Now that I remember, I think the problem with combining them into one repo is that there's no way to specify a root directory for a ruby gem, and the gem has a very specific and verbose file structure. I'd ideally want them to be in the same repo though, might have to investigate more.

msecret:
This is making my head hurt a little. My idea would be:
1. You checkout a release branch from master.
2. Update the version numbers, add, commit submit a PR for each repo (ruby, npm)
3. Have the code reviewed
4. If it's ready, tag the commit on the release branch.
4. Reviewer merges those changes into master.
5. Now a release build will be triggered because you merged into master with a new tag.

Is that consistent with what you were thinking? I think a whole other discussion is how could we do this all in one repo (ruby and npm). I'd really like that but not sure it's super easy to figure out.

jeremiak:
@msecret what do you think about merging it now and we're going to need a new release when #80 is closed so we can test it then?

PR #74, opened 2016-05-16
Add missing scss-lint plugins to stylelint

el-mapache:
See issue: https://github.com/18F/frontend/issues/131
- Adds stylelint plugins to fill in missing sass-lint rules.


msecret:
This is cool, and seems pretty easy to make plugins.

It'd be good to find a way to package up the rc file and plugins into a module so we could easily install it into projects.

el-mapache:
@marco Agreed. We'll need to make a separate repo then? To run stylelint with custom plugins we need to use the node API (as opposed to the command line tool). Given that, what do we want to expose to the module's consumer?

For example, stylelint has a number of packages you can install, one for their base lint config, one for each plugin, etc. I don't have a ton of experience authoring npm packages, is there a way to give a developer access to each of these things? Or, should we just expose a function that runs the linter as its configred here, and optionally accepts overriding options?

msecret:
I'm going to close this because we ended up making a separate package.

PR #77, opened 2016-06-02
Include release script and documentation update

jeremiak:
@msecret would really like your review of this. Releasing was kind of confusing and was many steps so I wrote a small release script to help us out. The release script does the following (lifted from `RELEASE.md`):

> ### How the release script works
> 
> Automation is important, and following checklists is hard so this script should do most of the work for you. `npm run release` simply calls `/scripts/release.js`.
> 
> The script does the following things in order:
> 1. Prompt you for the new release version number and main idea.
> 2. Update the `package.json` with the version number from above.
> 3. Checkout the `master` branch and cut a release branch from there. The release branch name will be the version number preceded by "release-".
> 4. Commit the `package.json` version bump.
> 5. Tag that commit with a tag that is just the release version.
> 6. Push that branch and tag to Github.
> 
> It also handles updating [18f/cg-style-gem](/18f/cg-style-gem) by executing the following steps:
> 1. While on the release branch, build the gem by running `npm run gem`.
> 2. Update `lib/cloudgov-style/version.rb` with the release version number.
> 3. Cut a new release branch from the `master` branch of the gem repo. The release branch name will be the version number preceded by "release-".
> 4. Commit the `version.rb` version bump.
> 5. Tag the commit.
> 6. Push the branch and tag to Github.

My thought was that the release script would make sure the files were updated appropriately for a release and then push everything but the user would still have to open PR and detail the changelog, etc.

I also changed the way we clone the gem repo so that the script doesn't prompt the user for github creds.

Also note that this PR is against `ms-release_steps`, so if this is merged, then #73 also needs to be merged for this to be in `master`.


msecret:
The build has an error related to cloning the gem down, although I don't know if it's related to your changes?

jeremiak:
@msecret I think it could be related to my changes, since I changed the clone URL to use SSH instead of HTTPS and since Travis doesn't have keys and access to the repo it can't clone it even though its a public repo. Travis needs to use the HTTPS URL but the release script should use the SSH URL so that the user doesn't need to enter their Github credentials.

One potential solution (e87389f) is to have two different tasks, one called something like `gem-clone-https` and `gem-clone-ssh`. That way everything can use HTTPS but the release script can use SSH when it wants. What do you think?

msecret:
Yea i ran into this problem before, I think separate tasks is the best idea

jeremiak:
Now green and ready for your review @msecret

Issue #78, opened 2016-06-03
Adjust presentation of followup actions box


msecret:
In order to show users a cohesive form experience, form elements should be sized correctly relative to each other and the rest of the UI elements.

thisisdano:
I adjusted this issue to focus on all the elements of what I'm calling the `followup actions box` — the actions needed to complete an parent action, like adding a service instance.

---

In order to for the user to complete an action like adding a service instance, we need to clearly display the needed actions and next steps.

**Acceptance criteria**
- [x] Follow-up actions (FAs) should be clear and distinct from the existing page elements
- [x] FAs should fit in with the existing rhythm of the page
- [x] FAs should be cancel-able
- [x] It should be as clear as possible exactly what FAs are requested
- [x] It should be as clear as possible what parent action the FAs refer to

thisisdano:
This issue is addressed by two PRs:

`cg-style`: https://github.com/18F/cg-style/pull/86
`cg-deck`: https://github.com/18F/cg-deck/pull/338

msecret:
Note: I still have to implement code to make this cancellable

PR #86, opened 2016-06-10
New input and followup action styles

thisisdano:
**Why**
In order to for the user to complete an action like adding a service instance, we need to clearly display the needed actions and next steps. 

**What**
- created a new style to apply to an actions box
  - made the box display full width
  - made the color distinct from existing page elements
  - adjusted the display text styles to better highlight the specifics of the action (`strong.inline-block`)
- refined input styles to be closer to the rhythm of the page
  - reduced padding in the inputs
  - made input labels stronger
  - reduced size of selector dropdown indicator
  - modified spacing around inputs
  - added a "cancel" button style
  - added a NONFUNCTIONAL `cancel` button to the actions box

**Known issues**
- `cancel` button is nonfunctional; just here to show where it would go
- there is no way to dismiss the actions box
- users without the proper privileges can access the action button and actions box, but cannot complete the action. these users should not be exposed to actions they cannot perform
- these styles are not intended to be responsive

**Related Issue**
**Adjust presentation of followup actions box:** https://github.com/18F/cg-style/issues/78

---

**Before**

screen shot 2016-06-10 at 11 05 31 am

---

**After**

screen shot 2016-06-10 at 10 33 27 am


thisisdano:
Updated the bg color of the FAs box to be a bit more chill.

screen shot 2016-06-13 at 11 16 50 am

src/css/base/form.scss
```diff
@@ -11,14 +12,73 @@ legend {
...
+select {
+  background-size: 1rem;
+}
+
+fieldset div {
+  padding-top: .75rem;
+  padding-bottom: .25rem;
+}
+
+button.button-cancel {
```
msecret, line 61: Does this need to be qualified with `button` or can it just be `.button-cancel`?

src/css/base/form.scss
```diff
@@ -11,14 +12,73 @@ legend {
...
+  color: $color-base;
```
msecret: Can you put all these properties in alphabetically order?

src/css/components/modal.scss
```diff
@@ -0,0 +1,20 @@
+
+@import "../override_vars.scss";
+@import "../core.scss";
+
+// faux modal for now, but maybe real later?
+
+.actions {
+  padding: 3rem 2rem;
+  background-color: #DEEEEF; // DEEF!
```
msecret: Can we add this to `override_vars.scss`? Or use an existing color (preferred)? It can a color straight from the [standards](https://standards.usa.gov/visual-style/#colors).

src/css/components/modal.scss
```diff
@@ -0,0 +1,20 @@
+
+@import "../override_vars.scss";
+@import "../core.scss";
+
+// faux modal for now, but maybe real later?
+
+.actions {
```
msecret: I made something similar to this here: https://github.com/18F/cg-style/pull/88/files#diff-138dae25243f1dd1405b1ecacbb90189R7.

One difference in mine: I didn't want my box to expand to the full width of the page. I was thinking I could bring my stuff into this file with as something like `actions-small` or something.

src/css/components/modal.scss
```diff
@@ -0,0 +1,20 @@
...
+  padding: 3rem 2rem;
+  background-color: #DEEEEF; // DEEF!
+  margin: -2rem;
+}
+
+strong.inline-block {
```
msecret: Does this have to be qualified with `strong`? Is this something within the `.actions` div? if so, it might make sense to namespace it under that, such as `.actions-inline-block` or `.actions-whatever`.

src/css/components/actions.scss
```diff
@@ -0,0 +1,30 @@
...
+  border: 2px solid $color-gray-light;
+  clear: left;
+  display: block;
+  float: left;
+  padding: 1rem;
+}
+
+.actions-large {
+  padding: 3rem 2rem;
+  background-color: #DEEEEF; // DEEF!
```
msecret: I think this line can be removed

thisisdano:
Ah, whoops! I made that change and didn't save.

PR #87, opened 2016-06-10
Create distinction between sections and heds on app page [styles]

thisisdano:
See https://github.com/18F/cg-deck/issues/311 for full description.

Paired with https://github.com/18F/cg-deck/pull/339 in `cg-deck`

### DIFF
--- a/src/css/components/billboard.scss
+++ b/src/css/components/billboard.scss
@@ -39,3 +39,12 @@
     line-height: 2.4375rem;
   }
 }
+
+// testing
+
+.section-card {
+  border: 1px solid #bababa; // having trouble referencing $color-lightgray;
+  padding: 0 1rem;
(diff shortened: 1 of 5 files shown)


msecret:
This is currently blocked on finding a solution to the path problem in component files. I'll probably be looking into that issue.

src/css/components/billboard.scss
```diff
@@ -39,3 +39,12 @@
     line-height: 2.4375rem;
   }
 }
+
+// testing
+
+.section-card {
+  border: 1px solid #bababa; // having trouble referencing $color-lightgray;
```
msecret: You probably need to import the WDS defaults file: `@import "../../node_modules/uswds/src/stylesheets/core/_defaults.scss";`

PR #96, opened 2016-06-16
Fixed header accessibility

msecret:
See https://github.com/18F/cg-deck/issues/361.

@elainekamlley feel free to review if you have time.


msecret:
Accompanying fix in deck: https://github.com/18F/cg-deck/pull/378

jeremiak:
Looks good, merging now.

We'll want to cut a new release but we should be able to do that after #97 gets merged as well.

PR #98, opened 2016-06-24
Update code blocks to be scrollable

rogeruiz:
Because of the work on 18F/cg-docs#291, we need scrollable `<code>` blocks.


jeremiak:
This looks good to me! We'll need to do a release and publish to npm to get this reflect in cg-docs. @msecret can you and I pair on that today? I think I could handle it, but since you did one recently and there were some bugs I'd rather just pair on it and make sure it gets done right the first time :)

PR #104, opened 2016-07-05
Sidebar improved context, styles

msecret:
Adds styling to the sidebar to improve what context you're in and uses
different colors to show this. Adds styling for blue headers, different
entities besides the main nav in the sidebar, different colors for the
different levels and more visual cues to show what you're on.

Refs 18f/cg-deck#382

### DIFF
--- a/src/css/components/sidenav.scss
+++ b/src/css/components/sidenav.scss
@@ -1,14 +1,16 @@
 
 @import '../core.scss';
+@import '../override_vars.scss';
 
 $width-nav-sidebar: 300px;
-$color-grid-dark:   #727272;
-$color-grid-light:  #e3e4e5;
 $site-top:          124px;
 $color-light-font:  #ffffff;
-$color-level-1-bg:  #2c3e50;
-$color-level-2-bg:  #1f2b38;
-$color-level-3-bg:  #595959;
+$color-background:  $color-primary-darkest;
+$color-level-1-bg:  $color-background;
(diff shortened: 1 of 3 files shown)


thisisdano:
@msecret: I modified a bit of the style and template to clean up a couple inconsistencies with color, spacing, and highlighting.

A couple remaining things:
1. Selecting an org should load the page with the org's spaces, instead of selecting the "spaces" hed within the org. The "spaces" hed should only be a hed, not selectable.
2. When you're in an a space within a parent org, only the space should be the yellow bar highlight (only one yellow bar at a time) — the parent org should be unbold and blue, on its default background bg. See below:

screen shot 2016-07-06 at 2 22 39 pm

It would be nice to add the little icons in, but we're probably out of time right now, and I don't know how to do it!

msecret:
@thisisdano I think I'm having trouble understanding 1. For 2, that's definitely possible but might be a little difficult based on how the events and menu works. we should tackle is as a next step, is it a release blocker?

So for number 1, it sounds like it should do 2 things at the same time when clicked:
- Opens up that orgs menu, so you can see the spaces below
- Go to the orgs page on the right

Is that what you mean? Should the spaces link go anywhere/do anything?

thisisdano:
@msecret That's right — at least as conceived, the `Spaces` hed is just a header, like the `Organizations` header at the top of the sidebar.

PR #105, opened 2016-07-06
Improve USWDS alert styles

msecret:
Allows alerts to be a better color, not have a header text, use
our color for the icon, and not have top margin.


jeremiak:
I've checked out the docs site and the landing site and it looks like this is our first alert usage so it won't be a breaking change.

I'll go ahead and rebase the branch and merge.

PR #106, opened 2016-07-06
Updating package.json version to 0.7.0

msecret:


jeremiak:
This looks good to me, @msecret. You wrote the release so I'll let you go ahead and merge this when you publish the release notes.

Quick question on Screener: can we add the cg-landing and cg-docs repos so that we can catch potentially breaking changes across our whole inventory of sites?

PR #110, opened 2016-07-11
Fix sidebar padding

msecret:
These padding fixes ensure that the text in the links of the sidebar
will always be aligned in the left side in all states(hover, active,
etc).


jeremiak:
This looks good to me, I'll merge when green!

Issue #112, opened 2016-07-15
Spaces link on hover has incorrect padding


msecret:
I believe this is a regression where the padding on hover of the "SPACES" link is incorrect.

jeremiak:
Looks fixed to me

PR #113, opened 2016-07-15
Fix incorrect header hover padding

msecret:
Fixes by simply copying the padding to the hover rule as well as just
the a rule to make sure it's on both.

Ref #112


jeremiak:
Looks fixed to me

PR #117, opened 2016-08-04
Upgrade wds to npm version

msecret:
I think it should only update patch updates, we shoudl decide when to do minor updates because they are often breaking


msecret:
@juliaelman take a look

juliasolorzano:
Looks good and thanks for making those changes!

One more thing I found in relation to list items. You can see that there is a `margin-left` set at `0.9em` for all `ul` and `ol` items [here](https://github.com/18F/web-design-standards/blob/v0.10.0/src/stylesheets/elements/_list.scss#L3). There was recently [an update made](https://github.com/18F/web-design-standards/blob/staging/src/stylesheets/elements/_list.scss#L3) to remove this. Thinking it would be good to add a style to cancel this out for now, until another upgrade of the WDS is made to `cg-style`.

jeremiak:
@msecret Could you review my last few commits? I upgraded node and npm to be the same version of cg-dashboard so that `npm link` is easier to think about and get right.

src/css/base/lists.scss
```diff
@@ -0,0 +1,6 @@
+
+ul,
+ol {
+  margin: 1rem 0 1rem 0.9em;
```
juliasolorzano, line 4: Should we change this value to `@include margin(1em null);` to match the upcoming fix?

msecret, line 4: I'm keeping it as this because `@include margin(1em null);` breaks a bunch of things I don't want to break right now.

src/css/override_components.scss
```diff
@@ -8,12 +8,17 @@
     padding-right: 3rem;
   }
 }
-
+//
```
msecret: This should be removed.

PR #118, opened 2016-08-04
Fix docs sidenav

msecret:
The docs sidenav was broken because it had a different hierarchy then
then dashboard sidenav and wasn't tested first.


jeremiak:
Looks good to me

PR #119, opened 2016-08-05
Style for activity log component

jeremiak:
To provide the styles for 18f/cg-dashboard#533

### DIFF
--- a/documentation/_components/activity_log.md
+++ b/documentation/_components/activity_log.md
@@ -1,0 +1,55 @@
+---
+layout: component
+type: component
+title: Activity log
+---
+
+<ul class="activity-log">
+  <li>
+    <div class="activity-log-item">
+      <div class="activity-log-item-text">
+        The app crashed is text that might go here
+      </div>
+      <div class="activity-log-item-timestamp">
+        <span>Timestamp</span>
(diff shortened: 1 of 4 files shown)


documentation/_components/activity_log.md
```diff
@@ -0,0 +1,59 @@
+---
+layout: component
+type: component
+title: Activity log
+---
+
+<div class="activity-log">
```
jeremiak: @msecret What do you think about having the `<ul>` inside of an element with `.activity-log`? Do you think it should actually just be a class that is applied to the `<ul>` itself?

msecret: `<ul>` sounds fine for me. I've found that we don't have to be as prescriptive on cg-style as wds.

jeremiak:
@msecret This is now ready for review

src/css/components/activity_log.scss
```diff
@@ -0,0 +1,52 @@
+
+.activity-log {
```
msecret: It should technically be `activity_log` for the base part of the classname, because the Activity Log is one thing. The `-` means that it's a separate part, such as an element or modifier.

PR #124, opened 2016-08-11
Add style for left border highlights on activity feed

jeremiak:
Continuation of #119 to add [left border highlights](https://github.com/18F/cg-dashboard/issues/515#issuecomment-238965105)

### DIFF
--- a/src/css/components/activity_log.scss
+++ b/src/css/components/activity_log.scss
@@ -22,6 +22,10 @@
   width: 100%;
 }
 
+.activity_log-item-console p {
+  font-family: monospace;
+}
+
 .activity_log-item-error,
 .activity_log-item-warning,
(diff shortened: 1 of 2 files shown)


src/css/components/activity_log.scss
```diff
@@ -7,21 +12,38 @@
...
   align-items: center;
-  background-size: 2em;
-  background-repeat: no-repeat;
-  background-position-y: 40%;
   border-bottom: 1px solid #ccc;
   display: flex;
   flex-wrap: wrap;
   margin: 0;
-  padding-left: 2em;
+  padding-left: 1em;
```
msecret: Why are these `em` instead of `rem`? There's not font size set here, is there one higher up?

Issue #115, opened 2016-07-28
Implement atomic design in the styleguide


jeremiak:
To make the implementation of atomic design easier throughout cloud.gov projects by using Fractal, a component library framework.

Acceptance criteria:
- [ ]  As a user, I should be able to find and implement atomic designs from cloud.gov styles into my project.
- [ ] As a user, I should have sample code options available to me so that I can easily use cloud.gov atomic designs on my chosen web framework.
- [x] As a user, I should be able to visit a url to view these code samples and documentation.

juliasolorzano:
Work started here: https://github.com/18F/cg-style/tree/je-add_atomic_library

juliasolorzano:
@vz3 @thisisdano quick check-in to see what you all think the next steps should be in implementing the pattern library. We have the ground work ready for adding components. I made note of a possible pattern to add [here](https://github.com/18F/cg-design/issues/5#issuecomment-240788665) and would be great to get further direction for next steps 😄 

Also, [here is the current state of the pattern library](http://federalist.18f.gov.s3-website-us-east-1.amazonaws.com/site/18F/cg-style/) for your reference.

thisisdano:
The new pattern library should probably include the existing patterns from `cg-style` — that's probably a good place to start. Does anyone have a build of fractal running that includes all the USWDS? That would also be a good starting point — and helpful for others in the future.

thisisdano:
And, while I'm happy to use standard atomic taxonomy, I've also thought about how to organize atomic design here:

https://docs.google.com/a/gsa.gov/spreadsheets/d/1jz_R-92lRCUcOqzHBB9D-5AQvO6U7fZHJ_AKMChIkXM/edit?usp=sharing

Based on GE's Predix system: https://medium.com/ge-design/ges-predix-design-system-8236d47b0891#.rtdu7tkc9
And on stuff like Buzzfeed's low-level framework, Solid: http://solid.buzzfeed.com/
Which is, itself, based on basscss: http://www.basscss.com/

But, at the end of the day, it probably doesn't matter how everything looks and is organized, as long as it's there in some order or another. The flexible Solid-style, low-level framework–based design system is a pipe dream I have for USWDS, but I don't ever see it happening, realistically.

vz3:
Let's start with type hierarchy, palette, and logo for now. I would hold off on USWDS elements for now.

juliasolorzano:
[...]

PR #127, opened 2016-08-15
Creating a new pattern library for cg-style

juliasolorzano:
[Preview on Federalist](http://federalist.18f.gov.s3-website-us-east-1.amazonaws.com/site/18F/cg-style/)

This pull request is for the initial groundwork for a [Fractal](http://fractal.build/) based pattern library for the new styles for cloud.gov. It includes the 18f pages supported version of the previous style guide, [seen here](https://pages.18f.gov/cg-style/). This will be removed once the new style guide is launched.

NOTE: The `button` component and docs pages are meant as placeholders for now and are not meant to dictate the design decisions.

Refs https://github.com/18F/cg-style/issues/115


msecret:
One thing that makes me a little nervous is having compiled files checked into the repo in the docs folder. Wasn't this something we had to undo in WDS because of the merge conflict problems? Or is that not a concern? Are there any other concerns around this?

I was also confused why there was a lot of similar files in `docs/build` and `docs/assets`. Whats the reason for that?

juliasolorzano:
> One thing that makes me a little nervous is having compiled files checked into the repo in the docs folder. 

Because there is not a way to compile the files on Federalist now, we are needing to check-in the compiled files. However, @jeremiak has made me aware that he is working on a solve by adding the ability to run `npm run federalist` via build on Federalist. That said, just a reminder this is the initial work on the pattern library for us to build off of.

> I was also confused why there was a lot of similar files in docs/build and docs/assets. Whats the reason for that?

Yes. This is b/c running `npm run build-library` creates this build dir and compiles the files to run the site. Again, not optimal, but a place for us to start and build from! Open to other suggestions too.

msecret:
OK that sounds good then. When FEderalist allows running npm we can move to not having the compiled files in.

juliasolorzano:
@msecret right on - I will submit a separate pull request with that update once that solve is in Federalist!

msecret:
Does @jeremiak need to see this? How should we merge?

juliasolorzano:
@msecret since Jeremia and I paired on this work, my assumption was that he would not need to see it. Again, this is a starting ground for the initial work. IMHO it is totally up to you if it's okay to merge in.

PR #129, opened 2016-08-22
Fix linting errors

msecret:


msecret:
This will go into a quick `1.1.1` release to fix the heading font size rem fix.

jeremiak:
Nice, this looks good to me. Apologies for those linting errors, I keep forgetting to run lint. Would be great if lint errors broke a build just like they do on cg-dashboard

jeremiak:
Psh, you're way ahead of me: https://github.com/18F/cg-style/issues/130

PR #132, opened 2016-08-22
Switch to circle CI

msecret:
For consistency with dashboard and because travis has a lot of
availability and slowness problems.

PR #134, opened 2016-08-23
Fix sass complaint due to missing var

msecret:
Using the color var request importing all wds and local color
variables.


jeremiak:
Travis looks like its just freaking out and cannot connect to the network. Circle CI is also running `npm run lint` which is returning with no errors. I restarted the travis build to see if there is another issue, but if there isn't I'll merge when that build finishes as it is green on Circle and we're removing Travis anyway (#132)

jeremiak:
Travis PR build turned green, merging now.

PR #136, opened 2016-09-02
Cloud.gov search bar styles

juliasolorzano:
This pull request is the continuation of the work started in [this pull request](https://github.com/18F/cg-style/pull/64) (thank you @jcscottiii!). Below is the comp provided by @thisisdano on how the layout and interaction should function.

The integration of search for the docs is also [started here](https://github.com/18F/cg-docs/pull/252) and will be directly effected by the completion of this work.

cc0cee96-7026-11e6-8c01-741ac1cbc222


msecret:
We haven't been good about putting this is the instructions and reminding people about it, but did you run `npm run lint`?

src/css/components/search.scss
```diff
@@ -0,0 +1,56 @@
...
+  position: relative;
+  $width: $search-btn-width;
+  margin-left: 0;
+  margin-top: 0.5rem;
+  margin-right: 0;
+  margin-bottom: 0;
+
+  [type='search'],
+  .search-field {
+    -webkit-appearance: none;
```
msecret: What does this do again? There's also https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-appearance. Should we just do `appearance` and auto-prefixer will take care of it?

src/css/components/search.scss
```diff
@@ -0,0 +1,56 @@
...
+    padding-top: 0;
+    width: calc(100% - #{$width});
+
+    @include media($small-screen) {
+      width: calc(100% - #{width});
+    }
+  }
+
+  [type='submit'],
+  .search-submit {
```
msecret: Does this and `search-field` have to be nested under `.search-bar`?

juliasolorzano:
@msecret yup! There is also some refactoring of this code, as you outlined in https://github.com/18F/cg-style/pull/64 that needs to be done too - hence the WIP.

src/css/components/search.scss
juliasolorzano: @msecret just putting a note here, this is still a WIP pull request and will most likely refactor this code.

However yes, this should change to `appearance` and let auto-prefixer do it's magic ✨

src/css/components/search.scss
juliasolorzano: `type=submit` probably should, but `.search-submit` does not. I'll take that into account as I refactor the code for the new designs ^5

juliasolorzano:
And just want to note here, this also reinforces my dislike for [WIP] pull requests for a few reasons:

1) When should the pull request be reviewed? How do folks get notified?
2) Could this potentially cause issues for feature bloat?

Going to continue to try it out, as it's a typical workflow used around 18F. However, I am going to noodle on how we might make this workflow more efficient.

PR #137, opened 2016-09-02
New app page design

msecret:
New app page design with routes, app settings and services.

Based on mockups here: https://github.com/18F/cg-design/issues/5

### DIFF
--- a/src/css/components/panel.scss
+++ b/src/css/components/panel.scss
@@ -65,19 +65,15 @@ h4 + .panel {
   }
 
   form {
+    background-color: $color-background-light;
     margin: -1rem;
+    padding: 1rem;
     width: calc(100% + 2rem);
   }
 
   &:last-of-type {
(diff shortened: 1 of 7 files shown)


jeremiak:
[Whoops, made the comment in the wrong spot](https://github.com/18F/cg-dashboard/pull/568#issuecomment-245050499)

src/css/components/panel.scss
```diff
@@ -0,0 +1,92 @@
...
+    padding: 1rem;
+    width: calc(100% + 2rem);
+  }
+
+  &:last-of-type {
+    border-bottom: 2px solid $color-background-light;
+  }
+}
+
+.panel-column {
```
jeremiak, line 79: Not a huge fan of `*-column` classes but I think this is fine for now since we're still implementing and designing the panel organism

jeremiak:
Looking good!

PR #143, opened 2016-09-07
Update versioning documentation

msecret:
Based on a discussion last week about whether a minor color change should be a major version upgrade.


juliasolorzano:
👍

PR #145, opened 2016-09-12
Form error

msecret:
Adds a simple form error class that modifies some things from the
wds input error class.

Also defines two variables for warning and error color, which are
currently the same value.


thisisdano:
Well, there's nothing in here that looks out of the ordinary. As long as it exists for a reason, it works for me.

PR #152, opened 2016-09-21
Panel upgrades

msecret:
Changes typography in the panel to match new designs, adding colors,
etc.

Should be minor version change.

### DIFF
--- a/src/css/components/actions.scss
+++ b/src/css/components/actions.scss
@@ -64,7 +64,7 @@
   }
 }
 
-/* TODO refactor all actions */
+/* TODO refactor "actions" into own components */
 .actions-large {
   background-color: $color-secondary-lightest;
   margin: -2rem;
--- a/src/css/components/actions.scss
+++ b/src/css/components/actions.scss
@@ -17,19 +17,25 @@
     color: darken($color-primary, 20%);
     text-decoration: none;
   }
-
 }
 
 // Extra specificity due to WDS.
 .action {
   &.action-cautious {
-    color: $color-gray;
+    background-color: $color-gray;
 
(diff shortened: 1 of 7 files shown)


src/css/components/all.scss
```diff
@@ -2,6 +2,7 @@
 @import 'actions.scss';
 @import 'activity_log.scss';
 @import 'billboard.scss';
+@import 'confirm';
```
jcscottiii: `confirm.scss`

src/css/components/confirm.scss
```diff
@@ -0,0 +1,38 @@
+
+@import '../../../node_modules/uswds/src/stylesheets/core/defaults.scss';
+@import '../override_vars.scss';
+@import '../core.scss';
+
+.confirm {
+  width: 100%;
+}
+
+.confirm-nexto {
```
jcscottiii, line 10: `confirm-next-to`?

src/css/components/actions.scss
```diff
@@ -17,23 +17,54 @@
...
+      box-shadow: inset 0 0 0 2px $color-warning;
+    }
+  }
+
+  &.action-finish {
+    background-color: $color-finish;
+  }
 }
 
+/* TODO refactor all actions */
```
jcscottiii: refactor how so? or put a link to an issue here.
thinking about if another dev came in, would they know what to do

src/css/components/confirm.scss
msecret: It should be that because that would mean `to` is an element or modifier, wheresas "next to" is the modifier.

PR #157, opened 2016-09-29
Margins for different amount of panel actions

msecret:
Configures margins correctly for when there are different amounts
of panel actions. So layout will be correct with one action vs
multiple. Used the adjacent sibling selector to only apply margin
when two actions next to eachother.

### DIFF
--- a/src/css/components/panel.scss
+++ b/src/css/components/panel.scss
@@ -46,6 +46,15 @@ h4 + .panel {
       margin-right: 0;
     }
   }
+
+  .action {
+    margin-left: 0;
+    margin-right: 0;
+
+    & + .action {
+      margin-left: 0.5rem;
+    }
+  }
 }
 
 .panel-header {


src/css/components/panel.scss
```diff
@@ -46,6 +46,15 @@ h4 + .panel {
       margin-right: 0;
     }
   }
+
+  .action {
+    margin-left: 0;
+    margin-right: 0;
+
+    & + .action {
```
thisisdano, line 54: This is controlling spacing for adjacent buttons, yes?

msecret: Yea adjacent sibling. If there are two actions next to eachother.

msecret:
Dan O. Williams (PDX; east coast hours) [10:45]  
Commented in the code, but this is for spacing between adjacent buttons, yes? If so, looks good

PR #164, opened 2016-10-07
Update typography

msecret:
Had to make some tweaks to make the typography acceptable in current
design.

screenshot_2016-10-07_11-41-26
screenshot_2016-10-07_11-41-55
screenshot_2016-10-07_11-42-22
screenshot_2016-10-07_13-01-34


adborden (approved):
Good from my perspective, up to @thisisdano to approve  :)

thisisdano:
Still workin' on fixing the issues with this change — especially in the docs. Hope to finish tomorrow.

thisisdano:
Worked on docs and dashboard; now on to the landing page(s)

src/css/base/typography.scss
```diff
@@ -4,7 +4,23 @@
 body {
   color: $color-base;
   font-weight: 400;
-  line-height: 1.7333333;
+  line-height: $leading-body;
+  font-size: $base-type-px;
+}
+
+a {
+  text-decoration: none;
```
msecret, line 12: So would this mean that the only indication something is a link is a difference in color? I believe that might be an accessibility problem.

src/css/override_vars.scss
```diff
@@ -23,29 +23,89 @@ $color-secondary-lightest:   #deeeef; // DEEF!
...
 $font-mono:                  'Roboto Mono', 'Bitstream Vera Sans Mono', 'Consolas', 'Courier', monospace;
+
+$base: 8;
+$base-grid: #{$base}px;
+$base-rem: #{$base}rem;
+$base-type: $base * 2;
+$base-type-rem: #{$base * 2}rem;
+$base-type-px: #{$base * 2}px;
+
+$bx5: #{($base / 2 / $base-type)}rem;    // 4px
```
msecret: I'm not sure I like the name of these, mainly because I don't know what they mean. Can you rename to something more understandable?

src/css/override_vars.scss
```diff
@@ -23,29 +23,89 @@ $color-secondary-lightest:   #deeeef; // DEEF!
...
+$b3x: #{($base * 3 / $base-type)}rem;   // 24px
+$b4x: #{($base * 4 / $base-type)}rem;   // 32px
+$b5x: #{($base * 5 / $base-type)}rem;   // 40px
+$b6x: #{($base * 6 / $base-type)}rem;   // 48px
+$b7x: #{($base * 7 / $base-type)}rem;   // 56px
+$b8x: #{($base * 8 / $base-type)}rem;   // 64px
+$b9x: #{($base * 9 / $base-type)}rem;   // 72px
+$b10x: #{($base * 10 / $base-type)}rem; // 80px
+
+$sans-s1: (8rem / $base-type);    // 8px;
```
msecret: So for these variables, you're starting low and moving up in values which is the opposite of how HTML works. As in h1 is largest while here, s1 is largest. Is there a reason for that?

src/css/base/typography.scss
```diff
@@ -13,88 +29,151 @@ h1, h2, h3, h4, h5, h6 {
...
+  @include media($medium-screen) {
+    font-size: $serif-s5;
+  }
+}
+
+.page-dek {
+  color: $color-mediumgray;
+  font-family: $font-sans;
+  font-size: $sans-s5;
+  font-weight: 200;
```
msecret: I think this should be 400, as that's the lowest weight we have right now for the 18Franklin font.

src/css/components/actions.scss
```diff
@@ -6,12 +6,11 @@
...
 .action-link {
   color: $color-primary;
+  font-size: $sans-s2;
   font-weight: bold;
   text-decoration: underline;
 
-  &:visited {
-    color: $color-primary;
-  }
+  &:visited { color: $color-primary; }
```
msecret: Please keep the formatting to what it previously was as that's whats compliant with the CSS styleguide.

msecret (commented):
All the images you've changed out have solid white backgrounds. Is there a reason for that?

thisisdano:
re: bgs -- I was having trouble getting them to output properly, so I added BGs. It wasn't going to affect the layout, so I left em in.

src/css/components/actions.scss
thisisdano: what does this comment mean? about one-line rules? seemed like I was getting lint errors when single-rulers weren't a single line — but I really don't care. i'll do em however.

src/css/override_vars.scss
thisisdano: just what I like — closer to something like basscss

src/css/override_vars.scss
thisisdano: They mean increments of a base unit (in this case 8px) — `$bx5` is `base * 0.5` and `$b5x` is `base * 5` — it's a way to specify a common whitespace scale without doing rem calculations all day.

src/css/base/typography.scss
thisisdano: I think we also have 18Franklin Light, which is `200`

src/css/base/typography.scss
thisisdano: no, they have an underline, but it's a border-bottom — at least in the docs, gotta do more scoping/spec'ing to make it work properly in the dashboard

src/css/override_vars.scss
thisisdano: Actually, it may still be the opposite of basscss (whose shortcode I don't really like for type scale...) IDK, maybe it should go the other way... but I like to think bigger number equals bigger face. I'm just one man, tho!

thisisdano: Still irritates me in markdown that smaller heds get more `#`s and thus more visual weight

src/css/base/typography.scss
msecret: Actually I just checked, 18Franklin Light is `300`, https://github.com/18F/cg-style/pull/164/files#diff-e56fe3386704ccaf26352285f872ddb6R8

src/css/override_vars.scss
msecret: What if we called them `basex5` etc?

src/css/components/actions.scss
msecret: Yes, they shouldn't be one-line rules. Let me know what linting errors you're getting.

src/css/override_vars.scss
thisisdano: To elaborate further — I think of these variables as elements of size or scale, like shoe sizes or dress sizes (higher number is greater scale), rather than elements of hierarchy or rank, like headers or positions in a race (higher number is lesser rank).

They also directly correspond (same names, same sizes) to the styles and type scale objects in the Sketch file, which could make things (potentially) easier for development...

src/css/override_vars.scss
thisisdano: I'm happy to consider a different variable for whitespace scale, though.

Also, FYI, all the objects in my design comps will be spaced on multiples of 8px.

thisisdano: thinking about this more.

maybe `$gx` is a good prefix ("grid times") and `05` is the way to go for half scale (as in Buzzfeed's Solid)

so: 

`$gx05`, `$gx1`, `$gx2` etc...

thisisdano: yeah, I like that more

src/css/base/typography.scss
thisisdano: ah, cool

thisisdano:
@msecret I think I'm done with this pass of style tweaks. I've also opened the PRs referenced above to apply once these new styles take effect.

There's still a LOT to be done on the dashboard to fix stuff that looks strange, but I think that's a job for another branch

thisisdano:
**landing**

screen shot 2016-10-14 at 9 00 50 am

---

**docs**

screen shot 2016-10-14 at 9 32 58 am

---

**dashboard**

screen shot 2016-10-14 at 9 41 48 am

npm-debug.log
```diff
@@ -0,0 +1,43 @@
+0 info it worked if it ends with ok
```
msecret: @thisisdano this file needs to be removed

src/css/override_vars.scss
```diff
@@ -14,34 +15,101 @@ $color-primary-alt-darkest:  darken($color-primary-alt, 70%);
...
+$font-serif:                 'Georgia', 'Cambria', 'Times New Roman', 'Times', serif;
+$font-mono:                  'Roboto Mono', 'Bitstream Vera Sans Mono', 'Consolas', 'Courier', monospace;
+
+// grid
+$grid: 8; // unitless for some calculations
+$grid-px: 8px;
+$grid-type: $grid * 2;
+$grid-type-px: #{$grid-type}px;
+
+$gx05: #{($grid / 2 / $grid-type)}rem;    // 4px
```
msecret: What's `gx`? Can we use full words, these variable names are too confusing.

src/css/override_vars.scss
```diff
@@ -14,34 +15,101 @@ $color-primary-alt-darkest:  darken($color-primary-alt, 70%);
...
+$color-cream:     #f5f5f3;
+$color-background-light:     $color-cream;
+
+// typefaces
+$font-sans:                  '18Franklin', system, -apple-system, BlinkMacSystemFont, 'Roboto', 'Helvetica Neue', 'Helvetica', 'Arial', sans-serif;
+$font-serif:                 'Georgia', 'Cambria', 'Times New Roman', 'Times', serif;
+$font-mono:                  'Roboto Mono', 'Bitstream Vera Sans Mono', 'Consolas', 'Courier', monospace;
+
+// grid
+$grid: 8; // unitless for some calculations
```
msecret: What does this value actually stand for? What about the grid?

msecret: Same as the ones below.

src/css/override_vars.scss
```diff
@@ -14,34 +15,101 @@ $color-primary-alt-darkest:  darken($color-primary-alt, 70%);
...
+$gx9: #{($grid * 9 / $grid-type)}rem;   // 72px
+$gx10: #{($grid * 10 / $grid-type)}rem; // 80px
+
+// widths
+$nav-width:                  $grid-px * 75; // 600px;
+$site-max-width:            $grid-px * 130; // 1040px;
+$text-max-width:             36em;
+
+// type scale
+$sans-s1: (8rem / $grid-type);    // 8px;
```
msecret, line 89: I still think we should move to reversing these values, so low numbers have high values like we were talking about.

msecret (commented):
Also, I'm still confused with the png files you changed. Don't we want those to have transparent backgrounds?

src/css/override_vars.scss
thisisdano: this value stands for the base whitespace scale multiplier (i.e. 8px) — it may not have any effect on the columnar "grid"

src/css/override_vars.scss
thisisdano: I don't really agree (see reasoning above), but if you _must_ insist I will consider.

src/css/override_vars.scss
thisisdano: I outlined my reasoning above. It is `grid * multiplier` — what level of verbose-ness do you desire? `$gridx05` or `$grid-05` — feels bulky and unnecessary to me, esp for something entered very often, but at the end of the day I don't care that much.

thisisdano:
And I don't really care at this point if the PNGs have bgs or not. I can re-export them if it feels important to have them clear.

src/css/override_vars.scss
msecret: I would say either of those options work. I think the bulk is necessary here to ensure other developers that work on this code can easily understand these variables that are used throughout the library.

src/css/override_vars.scss
msecret: Can we call it something like `$base-multiplier` then? Or add a comment next to it to explain that?

src/css/override_vars.scss
msecret: Ah, I read your comment too fast, I'm fine with keeping this.

src/css/override_vars.scss
thisisdano: ok, I'm changing this to `$grid-base` and commenting the code

src/css/override_vars.scss
thisisdano: ok, changing these to `$grid-05`, `$grid-1`...

src/css/override_vars.scss
thisisdano: thanks, sorry for the hassle — i don't know why i'm so prickly and pedantic about some issues of naming, but i am!

thisisdano:
OK, fixed those transparent images, too!

PR #165, opened 2016-10-12
New cg-grid for the updated grid sizing

adborden:
Allow us to migrate to a new grid on opt-in bases. USWDS grid is still
available.

@msecret would love to pair with you on this to go over some questions that came up.

### DIFF
--- a/src/css/components/grid.scss
+++ b/src/css/components/grid.scss
@@ -1,0 +1,9 @@
+.cg-grid {
+  @include outer-container;
+
+  @for $i from 1 through 12 {
+    .cg-width-#{$i} {
+      @include span-columns($i);
+    }
+  }
+}
--- a/src/css/components/all.scss
+++ b/src/css/components/all.scss
@@ -5,6 +5,7 @@
 @import 'confirm.scss';
 @import 'footer.scss';
 @import 'four_up.scss';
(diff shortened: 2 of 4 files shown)


adborden:
@msecret I think this is ready for review, unless there's something specific to do to add it to the styleguide?

I tested this against cg-docs and cg-landing and they are still using the `usa-grid` as expected.

src/css/components/grid.scss
```diff
@@ -0,0 +1,9 @@
+.cg-grid {
```
msecret, line 1: Now that I think about it, we don't use `cg` prefix for anything else. Could it just be `grid`? Or do you think it's a better interface to have the `cg` prefix?

adborden: Hmm, good point. I'm generally not a fan of prefixes except for frameworks meant to be built on top of. Although cg-style is a framework, it's not meant to be built on, it just contains a bunch of shared stuff. I'll remove the prefix.

adborden:
Also, I noticed on small screens the disclaimer has no padding. There's padding provided by `.content` for the main body grid, but I'm going to get rid of that, I think the grid should handle that kind of thing.

msecret (approved):

PR #168, opened 2016-10-18
Update style for new cg-site design

thisisdano:
**Paired with https://github.com/18F/cg-site/pull/9**

This update modifies styles to bring cg-site home up to new lightweight design.

---

PR #169, opened 2016-10-20
New sidebar styles

thisisdano:
In bringing the background to the new, lighter level, I needed to redesign the sidebar to match. This  PR is paired with PRs in `cg-site` (https://github.com/18F/cg-site/pull/9 - for the homepage), `cg-docs` (https://github.com/18F/cg-docs/pull/475 -for docs), and `cg-dashboard` (https://github.com/18F/cg-dashboard/pull/685 - for the dashboard).

It builds off the additional style work in https://github.com/18F/cg-style/pull/168, so please check that out if you haven't already.

---

We've updated to new typefaces and we're at the beginning of updating the dashboard to the styles shown in our [Invision](https://gsa.invisionapp.com/share/838UY596S#/191691942_App-Overview). For reference, this is how that looks:

screen shot 2016-10-20 at 1 06 10 pm

---

I've updated docs and dashboard to match this basic look and feel:

**docs**

screen shot 2016-10-20 at 5 01 13 pm

cg-docs-responsive 2x

**Dashboard**

screen shot 2016-10-20 at 1 03 40 pm

---

I've also added `<section> <h1>`s to all the docs content to give them consistent titles, and I've modified their hed structure when the body heds were out of whack.

This new functionality and look-and-feel works pretty well in all my testing, but obvs it's never perfect. This is a baseline for further work, so I'd be happy to talk about what's here, adjust it now, and look to improve it over further iterations!

It crosses over into Skyporter world as well, so all are tagged here!

@msecret @vz3 @nikzei @berndverst @brittag @jameshupp


thisisdano:
Updated nav highlight to black for accessibility and consistency (checked against white, but not against the light bg!)

thisisdano:
I also think that the blue bar next to the hamburger (in mobile) might better show the doc path (i.e. something like `/introduction/overview/` in the example above, to match the nav path, but that's a more involved solution, since that path should _probably_ match the url path, too...)

cg-docs-responsive 2x

thisisdano:
Added a crisp new favicon for good measure!

screen shot 2016-10-20 at 4 16 19 pm

suprenant:
@thisisdano nice, had noticed the favicon was falling apart!

src/css/base/typography.scss
```diff
@@ -181,4 +201,22 @@ pre {
     line-height: $leading-body;
     display: inline;
   }
-}
\ No newline at end of file
+}
+
+pre {
```
msecret: These rules and the `.code` one above should go in `src/css/base/formatted.scss`

src/css/base/typography.scss
```diff
@@ -107,21 +119,19 @@ h5 {
...
-  font-size: $serif-s4;
+  font-size: $serif-s5;
   line-height: $leading-body;
-
-  @include media($medium-screen) {
-    font-size: $serif-s5;
-  }
 }
 
+.usa-content.content h1 + p, // scope to docs
```
msecret, line 126: I think we need a better way to scope to docs. I'll think of a good way, it'll probably be a specific content class we use on docs.

src/css/components/actions.scss
```diff
@@ -22,16 +20,12 @@
 
 // Extra specificity due to WDS.
 .action {
-  &.action-primary {
-    font-size: $sans-s2;
-  }
+  &.action-primary { font-size: $sans-s2; }
```
msecret: Can you format these to 3 lines?

src/css/components/footer.scss
```diff
@@ -14,3 +14,12 @@
     color: $color-white;
   }
 }
+
+.content footer {
```
msecret, line 18: @thisisdano  What is this for? Where is this being applied to?

src/css/components/logo.scss
```diff
@@ -3,5 +3,14 @@
 
 .logo {
   height: 41px;
-  width: 203px;
+  width: 144px;
+}
+
+.header-title a {
```
msecret: I'd prefer to use a separate class for this, not a `a` tag selector. Like `.logo-link`.

src/css/components/sidenav.scss
```diff
@@ -150,76 +146,123 @@ $color-heading:     $color-primary-alt-light;
...
+        color: $color-textblack;
+      }
+
+      &:before {
+        background-color: $color-textblack;
       }
     }
   }
+
+  .sidenav-active-dashboard {
```
msecret: Rather then naming classes after the different sites, like `dashboard`, `docs`, `site`, I'd prefer we name them more specific to what they are trying to accomplish. So in this case, it looks like we're changing the colors to black. Would `sidenav-active-black` or something like that make more sense?

suprenant:
This has been blocked pending design review for a couple days. @vz3 please review ASAP

berndverst:
Big 👍  from me! This looks worlds better for the documentation. So much cleaner and more professional.

Looks great from Skyporter perspective.

thisisdano:
I modified the logo to be more closely aligned with login.gov.
@vz3 — all cool to continue?

---

screen shot 2016-10-27 at 1 26 12 pm

vz3:
👍 here

src/css/base/typography.scss
```diff
@@ -107,26 +119,25 @@ h5 {
...
   line-height: $leading-body;
-
-  @include media($medium-screen) {
-    font-size: $serif-s5;
-  }
 }
 
-.page-dek {
-  color: $color-mediumgray;
+.usa-content.content h1 + p, // scope to docs
```
msecret, line 126: Use separate class for docs specific.

src/css/base/typography.scss
```diff
@@ -159,6 +174,29 @@ blockquote {
...
+  padding-top: $grid-2;
+
+  p {
+    font-family: $font-sans;
+    font-weight: 400;
+    line-height: $leading-display;
+  }
+}
+
+.quote-attrib {
```
msecret: Consider making all of these quote component.

src/css/components/logo.scss
```diff
@@ -3,5 +3,14 @@
 
 .logo {
   height: 41px;
-  width: 203px;
+  width: 201px;
+}
+
+.header-title a {
```
msecret: Don't use `a` here.

src/css/components/section.scss
```diff
@@ -64,6 +75,60 @@ $chevron-size: 4.865rem;
...
+
+  a,
+  a:active,
+  a:hover,
+  a:visited {
+    color: $color-red;
+    text-decoration: underline;
+  }
+
+  a:active,
```
msecret, line 114: Ensure this rule isn't copied from above.

src/css/override_components.scss
```diff
@@ -183,3 +159,16 @@ input:focus {
     width: $grid-4;
   }
 }
+
+.fa {
```
msecret: @thisisdano We're currently only using fontawesome on docs, are you suggesting we use it on dashboard too? I think that requires a larger discussion.

src/css/components/sidenav.scss
thisisdano: this can be deprecated (and returned to `sidenav-active`) as the functionality is now the same

src/css/override_components.scss
thisisdano: This should be removed — I was trying something that I don't think is going anywhere!

src/css/base/typography.scss
thisisdano: sure, I can do that — I was hesitant to over-specify, but I can make it so

src/css/base/typography.scss
thisisdano: the issue with this is that it's difficult to add special classes to the docs markdown, so it's more reliable to do this via adjacency

src/css/components/footer.scss
thisisdano: This is the footer that appears at the bottom of docs content panels. 

screen shot 2016-10-28 at 12 50 35 pm

src/css/components/logo.scss
thisisdano: Should I just add a class to that `<a>` element in the markup?

src/css/components/section.scss
thisisdano: This is an overwrite to prevent writing a separate `a, a:visited` rule for color, but I can write it that way if desired.

src/css/base/typography.scss
msecret: Ah thanks for pointing that out @thisisdano, I'll look at it and consider options.

msecret (approved):

PR #176, opened 2016-11-03
Continuous deployment for pattern library via 18f-pages

adborden:
Automatically deploy the `master` branch of the pattern library to 18f-pages.

https://pages.18f.gov/cg-style/


adborden:
Forgot to mention, there's an issue with the assets path, so for example the loading animation appears as a broken img. Want to make sure this works in both dev and on 18f-pages, but I'm not sure we want to hold off on merging this, as I think it's a great improvement as-is.

msecret (approved):

PR #179, opened 2016-11-08
Style adjustments post 2.0 launch

thisisdano:
Address some of the concerns here: https://github.com/18F/cg-design/issues/20


src/css/components/panel.scss
```diff
@@ -121,3 +121,17 @@ h4 + .panel {
   flex-shrink: 1;
   min-width: 25%;
 }
+
+.panel-new {
```
msecret, line 125: What's panel new? Is this a new layout for the panel?

src/css/base/lists.scss
```diff
@@ -0,0 +1,12 @@
+h1,
+h2,
+h3,
+h4,
+h5,
+h6,
+p {
+  & + ol,
```
msecret, line 8: Love these `+` rules

src/css/base/typography.scss
```diff
@@ -18,6 +18,22 @@ a {
   &:active {
     color: $color-blue;
   }
+
+  /* stylelint-disable selector-no-qualifying-type */
+  // needed to target <a> styles
+  &.link-textblack {
```
msecret, line 24: Does the rule need to be `a.link-textblack` or can it just be `.link-textblack` on its own? Or is the former being overridden by something else?

src/css/base/typography.scss
```diff
@@ -27,7 +43,7 @@ strong {
 
 // TODO docs-class
 p strong,
-.usa-content.content li strong {
+.content-text li strong {
```
msecret, line 46: I like the new class name

msecret (changes requested):
You have 1 linting error;

```
src/css/override_vars.scss
 9:33  ✖  Expected "#BE4900" to be "#be4900"   color-hex-case
```

Let me know if you want me to make changes to anything.

src/css/base/typography.scss
thisisdano: it gets overridden by the `a` rule

src/css/components/panel.scss
thisisdano: I think it will be — but I wasn't ready to redefine all the panels just now. too many ripples

src/css/base/typography.scss
msecret: Yea I figured, I hope to fix that later.

src/css/components/panel.scss
msecret: When I have the chance, I'm going to write a TODO to remove this when we finalize the panels, because I don't like having separate code paths for new and old things.

msecret (approved):

PR #181, opened 2016-11-10
outline type in logo

thisisdano:
The logo isn't showing in safari, and it looks like a problem with the typeface. I've converted the type to outlines and I think it's all OK now!

This should go out with the next release if possible.

cc @msecret

- - -

**before**

screen shot 2016-11-10 at 11 50 20 am

- - -

**after**

screen shot 2016-11-10 at 11 50 42 am


msecret (approved):

msecret:
We could potentially patch just this change in as a quick fix too, I'll look into that

PR #203, opened 2016-12-06
Icons for CF entities

msecret:
Such as space, app, org. Includes code that makes it easier to inline
icons rather then the block layout as done on cg-site. To inline an
icon, which is done on dashboard overview page, simply don't use the
icon container and place the icon within the text tag it's next to.

The inlined icon will keep the size of the text tag it's in as it's
height and width are set to 1em. The icon also has a min width/height
though as I found at too small of sizes of text the icon wasn't
rendering very well.

Some of these icons need additional work to receive the border they
have.

This PR addresses separating out fill icons vs regular (or existing)
stroke icons. This has to be done because some of our icons only work
as fills and vice-versa for strokes. The default is just the `icon`
class and is stroke. The `icon-fill` class is for fill icons and will
be colored differently.

### DIFF
--- a/src/css/components/count_status.scss
+++ b/src/css/components/count_status.scss
@@ -25,6 +25,14 @@
   }
 }
 
+.count_status-icon {
+  float: left;
+}
+
+.count_status-text {
+  float: right;
+}
+
 .count_status-ok,
 .count_status-started,
 .count_status-running {
--- a/src/css/components/icon.scss
+++ b/src/css/components/icon.scss
@@ -6,8 +6,10 @@
   width: 3.9rem;
 
   .icon {
+    height: 100%;
     max-height: 3.1rem;
     max-width: none;
+    width: 100%;
   }
 
   & + h3 {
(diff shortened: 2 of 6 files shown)


msecret:
screen shot 2016-12-06 at 3 33 29 pm

adborden:
Your PR description would be great to add as docs to the pattern library :)

Do the icons have to be created a certain way for this to work? In past experience, if the designer doing the export isn't careful, the icon won't align with the text baseline properly.

What's the difference between these icons and the [src/img/*-small.svg](https://github.com/18F/cg-style/blob/master/src/img/app-small.svg) icons? They are obviously different when you look at the markup, but I'm not sure why.

msecret:
I was thinking about adding the icons to the pattern lib, so definitely want to try that out.

So there is some things I had to do with the icons svg file unfortunately. I had to remove some inline css styles from the elements and maybe set some classes. I'm looking to find ways to make the export more automatic, but might need to add some documentation if there's no way to do that.

Also, I'm waiting on new icons from @thisisdano , maybe we can work out best way to export.

msecret:
I actually didn't see the `small` icons. One thing: in the cg-design repo, the icons were exported by size, but since they are svg, they really shouldn't have different sizes. Maybe I should get rid of the app/space/org-small icons? Or are they used somewhere?

adborden:
Quoting Marco Segreto (2016-12-07 09:52:33)
> I actually didn't see the `small` icons. One thing: in the cg-design repo, the icons were exported by size, but since they are svg, they really shouldn't have different sizes. Maybe I should get rid of the app/space/org-small icons? Or are they used somewhere?

They are used in the breadcrumb nav prototype, so they are in
[cg-style](https://github.com/18F/cg-style/blob/master/src/css/components/nav-breadcrumb.scss)
but not in the dashboard.

--
Aaron D Borden
Human and Hacker

adborden:
Quoting Marco Segreto (2016-12-07 09:50:54)
> I was thinking about adding the icons to the pattern lib, so definitely want to try that out.
> 
> So there is some things I had to do with the icons svg file unfortunately. I had to remove some inline css styles from the elements and maybe set some classes. I'm looking to find ways to make the export more automatic, but might need to add some documentation if there's no way to do that.
> 
> Also, I'm waiting on new icons from @thisisdano , maybe we can work out best way to export.

FYI, @noahmanager did some exploration using gulp-sketch[1] to export icons
from sketchfiles as part of automating the icons in SCSS, but no code
was written for the SVG export (at least I couldn't track anything
down).

[1] https://www.npmjs.com/package/gulp-sketch

--
Aaron D Borden
Human and Hacker

PR #204, opened 2016-12-07
Bordered icons and documentation

msecret:
Documented icons and fill icons. Added a way for fill icons to be bordered. Refactored both icon sets to make a little more sense in terms of implementation, such as requiring fill icons needing the fill class to be color styled.


adborden (approved):
Very detailed, well done.

PR #205, opened 2016-12-08
Adding styles for routes panel on app page

line47:
Added styles for routes panel binding and unbinding. 
Created panel styles for routes and services.

### DIFF
--- a/src/css/components/confirm.scss
+++ b/src/css/components/confirm.scss
@@ -36,8 +36,8 @@
   font-weight: 600;
   margin-right: 1rem;
 
-    a {
-      color: $color-warning;
-      text-decoration: underline;
-    }
+  a {
+    color: $color-warning;
+    text-decoration: underline;
+  }
 }
--- a/src/css/components/confirm.scss
+++ b/src/css/components/confirm.scss
@@ -19,9 +19,9 @@
 .confirm-over {
   align-items: center;
   border: 0;
-  display: flex;
   justify-content: space-between;
-  padding: 0;
+  padding: 0.5rem;
+  background-color: #f5f5f3;
 }
 
 .confirm-block {
@@ -35,4 +35,9 @@
   color: $color-warning;
   font-weight: 600;
   margin-right: 1rem;
+
+    a {
+      color: $color-warning;
+      text-decoration: underline;
+    }
 }
--- a/src/css/components/confirm.scss
+++ b/src/css/components/confirm.scss
@@ -18,10 +18,11 @@
 
 .confirm-over {
   align-items: center;
+  background-color: #f5f5f3;
   border: 0;
   justify-content: space-between;
   padding: 0.5rem;
-  background-color: #f5f5f3;
+
 }
 
 .confirm-block {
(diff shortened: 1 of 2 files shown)


src/css/components/confirm.scss
```diff
@@ -19,9 +19,9 @@
 .confirm-over {
   align-items: center;
   border: 0;
-  display: flex;
   justify-content: space-between;
-  padding: 0;
+  padding: 0.5rem;
+  background-color: #f5f5f3;
```
msecret: These should be alphabetized by property name so:

```
  align-items: center;
  background-color: #f5f5f3;
  justify-content: space-between.
 ...
```

src/css/components/panel.scss
```diff
@@ -157,3 +156,23 @@ h4 + .panel {
   font-weight: 600;
   padding-bottom: $grid-1;
 }
+
+.panel-routes,
```
msecret: This code makes sense and works, but I'm worried we have way too much in the panel component. I also am unsure we should be writing classes as specific to one use case, like `panel-services`.

I guess the question to ask is, why are we modifying these things in our heading tags just for panel services and how could we use classes and styles that already exist (there's not an easy answer to this). For instance, where does the 0.75rem style come from, is there not another `h` tag we can use for that?

src/css/components/panel.scss
```diff
@@ -157,3 +156,23 @@ h4 + .panel {
...
+    font-weight: 400;
+  }
+
+  h3 {
+    font-size: 0.75rem;
+
+    a {
+      color: #056dd4;
+
+      .route-link {
```
msecret: I wrote code recently so that all links in these panels will be underlined, so I think this is not needed.

src/css/components/panel.scss
```diff
@@ -157,3 +156,23 @@ h4 + .panel {
...
+
+  h1 {
+    font-weight: 400;
+  }
+
+  h3 {
+    font-size: 0.75rem;
+
+    a {
+      color: #056dd4;
```
msecret: What's this color from? I think all our colors should use variables from `override_vars.scss`. So this would be `color: $color-primary` or something.

I also question again why we can't just use the default styling of links.

msecret (changes requested):

src/css/components/confirm.scss
line47: Reordered

src/css/components/panel.scss
line47: removed panel specific classes

src/css/components/panel.scss
line47: Removed

src/css/components/panel.scss
line47: Removed

src/css/components/confirm.scss
```diff
@@ -18,10 +18,11 @@
 
 .confirm-over {
   align-items: center;
+  background-color: #f5f5f3;
```
msecret: We shouldn't be using straight color values anywhere, everythign should come from the variables defined in `override_vars`.  So this should be `background-color: $color-lightestgray;`

msecret (changes requested):

msecret (approved):

PR #209, opened 2016-12-10
Change table row layout

msecret:
So that things are more right aligned execpt first column, there's more padding, and the lines are consistent.


msecret:
Checked cg-site, things look fine and Dan said it's good to go.

line47 (approved):
The padding, borders and text alignment look good.

PR #212, opened 2016-12-12
Include library assets in the watch command

adborden:
This combines some work that I and @thisisdano did around improving the workflow for building the library. The watch command now includes copying the assets for the library and breaks down the watch command into separate watchers so that the watch builds are a bit faster. To use, it should be the standard workflow:

```
# first terminal, watch the library itself
$ npm run start

# second terminal, watch the js/styles/assets
$ npm run watch
```


msecret:
Does this fix this: https://github.com/18F/cg-style/pull/210

msecret (approved):

PR #213, opened 2016-12-13
Make panel row clickable

msecret:
Makes the panel row clickable. I used negative margin to pull out the padded row while keeping the same total size. I also added a pointer cursor.


line47 (approved):
I think this looks good.

PR #214, opened 2016-12-13
Removing nested P from activity_log-item_text

line47:
Removing the nested P as the class name has shifted to the P element.
See https://github.com/18F/cg-dashboard/pull/809


line47:
It looks like those changes have already been made and merged. Closing this.

PR #215, opened 2016-12-13
Added style for bordered panel row and lighter panel title weight

line47:
Added style for bordered panel row and lighter panel title weight.

### DIFF
--- a/src/css/components/panel.scss
+++ b/src/css/components/panel.scss
@@ -8,7 +8,7 @@ $panel-padding: 1rem;
 .panel {
   @include clearfix;
   min-height: 10rem;
-  padding: 0.75rem 0 0 0;
+  padding: .75rem 0 0;
 
   & + .panel {
     margin-top: 1.5rem;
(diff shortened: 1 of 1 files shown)


src/css/components/panel.scss
```diff
@@ -24,12 +24,20 @@ h4 + .panel {
 
 .panel-title {
   font-size: 1.3rem;
-  font-weight: 600;
   margin-bottom: 2rem;
+
+  h1,
```
msecret: Currently, panel title is applied directly to the `h` tag, so this shouldn't be needed.

```
<h1 className={ this.styler('panel-title') }>{ this.props.title }</h1>
```

src/css/components/panel.scss
```diff
@@ -80,6 +91,12 @@ h4 + .panel {
     margin: 1rem -1rem -1rem;
     padding: 1rem;
     width: 100%;
+
+    .route-form-preview {
```
msecret: This should actually go in the dashboard code:

https://github.com/18F/cg-dashboard/blob/067ef8fe56532ed2e032a18be9ad53b357d11586/static_src/css/route_form.css

I mean, technically it should go in cg-style, but we don't have a place for it yet, so it would do better there for now.

msecret (changes requested):

msecret (approved):

PR #223, opened 2016-12-15
Use npm auth token for npm publish on CI

adborden:
I've set the NPM_API_TOKEN env variable in CircleCI with the 18f-devops token.

This avoids
- having to put a user/password in CircleCI.
- risk of leaking secrets when piping input around.


adborden:
... and you can get an API token by logging in with `npm login`, then `cat ~/.npmrc`.

msecret (approved):

PR #225, opened 2016-12-16
Pattern Library upgrades

thisisdano:
OK — there's a lot of stuff in here, too much I know, but it all got wrapped up together and now it's gotta get merged.

What's here:
- new 10-step type scale
- new 18Franklin v22
- all new styles for the fractal instance
- new organization of pattern library

screen shot 2016-12-16 at 8 39 49 am

- added base patterns from Sketch to pattern library
  - colors
  - type scale
  - typefaces
  - line-height
  - whitespace scale
  - opentype features for 18Franklin
  - layout grid
  - border options
- fractal template upgraded to nunjucks templates from handlebars for better logic and readability (this is what the FEC fractal pattern library uses)
- developed a set of single purpose utility classes (a la tachyons) for prototyping and as a layout helper
- new component build for icons [**breaks current layout**]
- new outline style for icons
- full set of square fill icons (including reverse icons for almost all needs)

screen shot 2016-12-16 at 8 33 24 am

- developed style for `trigger-buttons` i.e. the buttons that trigger dropdowns (as in the breadcrumb)
screen shot 2016-12-16 at 8 34 43 am

- new optional flex grid that can be used a layout helper
- pattern styling for panels
- updated component build for overview org quicklook cards [**breaks current layout**]

- - -

All these changes constitute a starting point for this last development push, and any continued development post-Navigator.  I know there's a lot to see here, and a few component layout wrinkles to straighten out before we can merge, but I want to work to get this in ASAP.

ONWARD!

@line47 @adborden @msecret 

**Requested changes in this PR**

- [x] [remove `background-color` and `color` from `.grid-container`](https://github.com/18F/cg-style/pull/225#discussion_r93670429)
- [x] Resolve confusion about stroke/fill icons, is `.icon-fill` used only for
  fill icons? How do stroke/fill icons relate to `icon-ui`, `icon`,
  `icon-square`? **[seem like a separate issue]**
- [x] remove `&-modifier` based rules, just use the expanded name [[source](https://github.com/18F/cg-style/pull/225#discussion_r93310717)]
- [x] rename card-home-contents to card https://github.com/18F/cg-style/pull/225#discussion_r93309845 **[changed style names, not feature name, as the feature is that specific card]**
- [x] make it clear in the docs that the utility classes are meant for
  prototyping, not for use in production
- [x] documentation on the `0` classes
  https://github.com/18F/cg-style/pull/225#discussion_r93677892 **[0 classes are an error, they should be `--` classes — still, could use docs, but necessary for this PR?]**
- [x] drop the clearfix, which should exist in wds
  https://github.com/18F/cg-style/pull/225#discussion_r93679802

**Requested changes for a new PR (or multiple Issues/PRs)**

- [ ] add helper text to action components and realistic action text content https://github.com/18F/cg-style/pull/225#discussion_r93788438 **[not sure this is a clear benefit]**
- [ ] rename `row` to `col-`
  https://github.com/18F/cg-style/pull/225#discussion_r93314927 **[I don't see this one, and I don't agree that `row` should be `col`]**
- [ ] clear documentation for utility classes
  https://github.com/18F/cg-style/pull/225#discussion_r93312790 **[necessary for merge?]**
- [ ] clarify use of `@extend` in CONTRIBUTING.md
- [ ] more complete documentation for components like button-trigger, icon-ui
- [ ] use sass `map` for scale labels/values
  https://github.com/18F/cg-style/pull/225#discussion_r93677552 **[I will try and do this, but consider nice-to-have and unnecessary for merge]**
- [ ] resolve conflicts/naming of utility classes
  https://github.com/18F/cg-style/pull/225#discussion_r93680792 **[I don't see conflicts rn, need specific examples]**
- [ ] include documentation on component status
- [x] documentation on copy/print vocabulary **[changed to `ll` (line-length) for consistency]**
  https://github.com/18F/cg-style/pull/225#discussion_r93678147
- [ ] card contents should have an preview demonstrating open (it's not working?) https://github.com/18F/cg-style/pull/225#discussion_r93673806 **[agree in principle, but don't see a reason to tie it to this PR, since we're not even using this card code in production]**
- [x] replace reset.css with npm version
  https://github.com/18F/cg-style/pull/225#discussion_r93683817 **[removed, unnecessary rn]**
- [ ] Mark breadcrumb nav as "Prototype" status and reduce the 3 separate files full of duplicate code https://github.com/18F/cg-style/pull/225#issuecomment-269049790 **[already marked as wip]**
- [x] Mark the overview component as "Prototype" or change it's class structure so it doesn't rely on atomic classes https://github.com/18F/cg-style/pull/225#issuecomment-269049790 **[already marked as wip]**
- [x] new status names for pattern status https://github.com/18F/cg-style/pull/225#issuecomment-269038074


line47:
I think it would help to have this broken down into a 4 separate smaller PRs for the following:
- Fonts
- SVGs
- Pattern Library files
- SCSS updates

thisisdano:
@line47 I agree, but I don't know how to do that!

thisisdano:
However, since they're all so interrelated, it may just make sense to try and evaluate it using those buckets, but to leave the whole thing together.

line47:
The only styling issue I am seeing when running this locally is icon sizing and that has been fixed in https://github.com/18F/cg-dashboard/pull/853

Icon issue screenshot
screen shot 2016-12-19 at 2 57 49 pm

adborden:
There are some advanced tools to help reorganize this into multiple PRs, namely `git rebase -i` for git's interactive rebase. But it's not for the faint of heart. You might end up in a worse-off place than you started if you're not familiar with rebasing. It's a big PR, but let's just not add anymore to it.

The easiest thing to do is to look linearly through the commit history and pick good checkpoints to merge and create a new PR based from there. Then a new PR from the merge to the next good checkpoint commit.

docs/src/main.css
```diff
@@ -30,7 +30,7 @@ code {
 }
 
 .grid-container {
-  background-color: #ddd;
+  background-color: white;
```
adborden: I'm not sure this is what we want, white text on white background. Also, is this different from `$color-white`?

adborden:
Another tip, I'm going through commit by commit, because it's easier to review these large PRs change by change and I'm seeing a lot of commits where files are created and then moved or deleted, or variables that don't exist in the final PR. Sometimes it makes sense  to ["squash" several subsequent commits](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) into a single one. I do this a lot.

thisisdano:
@adborden Yeah, my commit style SUCKS because I don't know what I'm doing. Going through commit by commit will probably make you crazy. I'm sorry for that

thisisdano:
Everything rebase related totally spooks me :(

docs/src/components/component/icon-square/icon-square.html
```diff
@@ -0,0 +1,6 @@
+
+<svg class="icon-{{ iconSize }} icon-fill {% if isBorder %}icon-bordered {% endif %}{% if iconStatus %}icon-{{ iconStatus }}{% endif %}">
```
adborden: I thought there was a distinction between fill icons and stroke icons, have we converted all the icons to fill icons?

src/css/components/icon.scss
```diff
@@ -24,16 +24,27 @@
...
+    height: $grid-2;
+    width: $grid-2;
+  }
+  &-medium {
+    @extend .icon;
+    height: $grid-3;
+    width: $grid-3;
+  }
+  &-large {
+    @extend .icon;
```
adborden: Our [CONTRIBUTING guide](https://github.com/18F/cg-style/blob/master/CONTRIBUTING.md#build) says to avoid `@extend` (which I don't really agree with) but if that's the case let's talk about it here or document the exception.

docs/src/components/component/icon-ui/README.md
```diff
@@ -0,0 +1,72 @@
+
```
adborden: This file exists twice, here and in the [icon-square component](https://github.com/18F/cg-style/pull/225/files#diff-1f200b905f2f27aabf16cf064b1a7bc9). Probably only one of them is needed?

src/css/override_components.scss
```diff
@@ -97,12 +97,32 @@ button,
...
+    box-shadow: inset 0 0 0 1px $color-blue;
+
+    &:hover {
+      background-color: $color-white;
+      box-shadow: inset 0 0 0 1px $color-textblack;
+      color: $color-textblack;
+    }
+  }
 }
+/* stylelint-enable */
```
adborden: Would you please move this up a line so it matches the indent/block where the rule is disabled?

docs/src/components/component/button-trigger/button-trigger.config.yml
```diff
@@ -0,0 +1,129 @@
...
+  - name: disabled-outline-open
+    context:
+      isOutline: true
+      isOpen: true
+      buttonStatus: cautious
+      buttonStyle: outline
+      buttonText: disabled-outline
+  - name: finish-outline-open
+    context:
+      isOutline: true
```
adborden: Can we match the language we're using in the dashboard? For example, the [buttons have a "type"](https://github.com/18F/cg-dashboard/blob/master/static_src/components/action.jsx#L16), of which "outline" is one possible value. I suggest standardizing on "kind" like you're using elsewhere in the PR for similar concepts.

docs/src/components/component/button/button.html
```diff
@@ -0,0 +1,11 @@
+{% if isDefault %}
+<button class="action action-{{ buttonStatus }}">{{ buttonText }}</button>
```
adborden: In cg-dashboard we're blurring the lines between Button and Action. I suggest we rename this component to "Action" to be consistent. I'm not sure if the difference actually warrants separate components though.

docs/src/components/component/button/button.config.yml
```diff
@@ -0,0 +1,125 @@
...
+      isLink: false
+      buttonStatus: cautious
+      buttonText: disabled
+  - name: finish
+    context:
+      isDefault: true
+      isOutline: false
+      isLink: false
+      buttonStatus: finish
+      buttonText: finish-outline
```
adborden: I feel like this text isn't a ton of value. I would prefer to have real-ish text in here. Or if you really want this kind of text, maybe we can create it from the other variables without having a dedicated one.

src/css/components/actions.scss
```diff
@@ -25,44 +24,104 @@
...
+      color: $color-finish;
+
+      &:hover {
+        box-shadow: inset 0 0 0 1px $color-textblack;
+        color: $color-textblack;
+      }
+    }
+  }
+
+  &.action-base {
```
adborden, line 98: When would I use `action-base` over `action` or `action-primary`?

adborden (commented):
Phew! I haven't gotten through all of it, but it's looking pretty good from my perspective. Thanks for all the hard work!

Overall, I'd like to see us go for depth rather than breadth of work. I would like to see more documentation and complete components rather than implementing every single existing component in templates. For me, I want to know _when_ to use one component over another. The templates cover the _how_, but we're still missing the when. I think docs are part of this, but also using more concrete examples with the templates.

There's also still some confusion for me over the icons. We have `icon`, `icon-ui`, and `icon-square`, and I don't really understand why. Reading the docs, I thought the distinction between fill and stroke icons was important, but I don't see that reflected in the pattern library.

docs/src/main.css
thisisdano: First, thanks for accepting this trial of Hercules! Now, on the comments!

Dunno why this is like this — I inherited it. 
Colors in the main.css file in the docs aren't run through a sass preprocessor, so we can't use the sass variables. Otherwise yeah, I try to use vars everywhere possible. 

THAT SAID — this is only the stylesheet for the PL, so it's pretty sloppy and low priority!

docs/src/components/component/icon-square/icon-square.html
thisisdano: No — there are still some stroke icons (used on the marketing home, but that may change)

src/css/components/icon.scss
thisisdano: Yeah, I've noted that and understand the impulse — extends can tend to mess with the cascade and don't work across a media query boundary. However, within limits, they seem pretty reasonable — for neighboring elements in the sheet, those that don't involve media queries, and those that have no rule overlap. Should they be used here? IDK — I could always rewrite them with a copy/paste, or assume the designer would use both classes.

docs/src/components/component/icon-ui/README.md
thisisdano: Yeah, that's just sloppy setup work on my part.

docs/src/components/component/button-trigger/button-trigger.config.yml
thisisdano: Yeah, this is part of a larger question about how important consistency is in the docs and how closely the PL patterns can/should match the production React patterns.

I added the boolean values here (like `isWhatever`) before I upgraded to the nunjucks template — handlebars couldn't handle "complex" logic like == matching.

docs/src/components/component/button/button.html
thisisdano: Interesting! Feels strange to kill the `Button` but it makes sense.

docs/src/components/component/button/button.config.yml
thisisdano: I kinda understand, but when the buttons are collected on the same page without labels, it can be hard to know their name. Now, this is maybe a problem that we could solve with better/different preview templates (maybe) but for now I liked being able to know the name of the component variant at a glance on a collated page.

src/css/components/actions.scss
thisisdano: The one place I can think of where there's a base action state is in the breadcrumb nav, where the trigger buttons start in a base state.

line47:
The font work that was part of this PR was merged earlier today: https://github.com/18F/cg-style/pull/227

docs/src/components/base/borders/borders--2px.html
```diff
@@ -0,0 +1,7 @@
+{% for color in colors %}
```
msecret, line 1: Can we try and keep the naming patterns for files consistent? This can be the same as our variable naming scheme described in the [front end guide](https://pages.18f.gov/frontend/#suggested-custom-methodology) and our contributing doc.

So they should all be `borders-4px.html`, there shouldn't be any BEM style `--`s.

docs/src/components/component/button/button.html
msecret: I'd agree with Aaron, we can kill the `button`.

docs/src/components/component/icon-square/icon-square.html
msecret: Yea we weren't able to convert them, maybe it should be some future goal, but I'm not sure it's possible.

src/css/components/card.scss
```diff
@@ -0,0 +1,27 @@
+@import '../../../node_modules/uswds/src/stylesheets/core/defaults.scss';
+@import '../override_vars.scss';
+@import '../core.scss';
+
+.card-home-contents {
```
msecret: I wan this to be more generic then for "home". If it's not being used anywhere else, why can't we just call it `card`?

src/css/components/grid.scss
```diff
@@ -24,3 +24,82 @@
     }
   }
 }
+
+/* stylelint-disable declaration-no-important  */
+%row {
```
msecret: The amount of `!important` in this file sounds like a bad idea. By looking at it, I can't really see why there's a need for important.

src/css/components/icon.scss
msecret: This isn't in the styleguide, because when I wrote it, it wasn't a commonly used thing in scss, but I also don't think I want to use the `&-name` functionality in scss because it makes things hard to search for. I want to be able to search for `icon-small` in the whole codebase and find it.

In terms of the `@extend`, I would expect you'd have to use both `icon` and `icons-{size}` classes to get an icon to render based on size (which is how most of the exiting code works. So I'd prefer not extending and requiring devs to use both classes.

I also had separate comments on having icons sizes, I think we resolved that here: https://github.com/18F/cg-dashboard/pull/853.

src/css/components/icon.scss
```diff
@@ -61,8 +82,32 @@
...
+  }
+}
+
+.icon-chevron {
+  fill: $color-blue;
+  height: 11px;
+  width: 11px;
+
+  &-left {
+    @extend .icon-chevron;
```
msecret, line 95: Same comments as above.

src/css/components/icon.scss
```diff
@@ -61,8 +82,32 @@
...
   }
 
-  &.icon-alt {
-    fill: $color-primary-alt-light;
+  &.icon-active {
+    fill: $color-blue;
+  }
+}
+
+.icon-chevron {
```
msecret, line 89: So in the pattern library, this is called `icon-ui`. Can we keep the naming conventions consistent? I'd prefer this to be `icon-ui`, but I'll leave it to you what you think it should be based on how you think it will be used.

If it's a separate component in the styleguide, it also might make sense to be it's own component file.

src/css/components/trigger.scss
```diff
@@ -0,0 +1,74 @@
+@import '../core.scss';
+
+.usa-button.trigger {
```
msecret, line 3: I'm having trouble understanding what `button trigger` is even after viewing styleguide. It looks like a sort of drop down button? 

I also think the styleguide should be consistent and should be called trigger.

src/css/override_vars.scss
```diff
@@ -77,47 +82,73 @@ $grid-7:             #{($grid-base * 7 / $grid-type)}rem;     // 56px
...
+$lh-hed:       1.05;
+$lh-cap:       1.3;
+$lh-body:      1.55;
+$lh-display:   1.65;
 
 // border radius
-$rounded:           2px;
-$not-rounded:       0;
+$rounded:         2px;
+$rounded--:       0;
```
msecret: I'd prefer something more understand the `--`. I would not understand that as a developer.

src/css/override_vars.scss
```diff
@@ -57,6 +57,11 @@ $font-sans:                     '18Franklin-webfont', system, -apple-system, Bli
 $font-serif:                    'Georgia', 'Cambria', 'Times New Roman', 'Times', serif;
 $font-mono:                     'Roboto Mono', 'Bitstream Vera Sans Mono', 'Consolas', 'Courier', monospace;
 
+// utility class scale suffixes
+
+$u-breakpoints: '-xs', '-s', '-m', '-l', '-xl';
+$u-flexscale: '-f1', '-f2', '-f3', '-f4', '-f5', '-f6', '-f7', '-f8', '-f9', '-f10', '-f11';
```
msecret: I'd prefer more understandable. As a developer, how will I know `f` stands for `flux` and not something else.

src/img/i-fill-app--reverse.svg
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
msecret, line 1: Are these reverse icons being used anywhere? I couldn't find them in style guide.

msecret (changes requested):

src/img/i-fill-app--reverse.svg
thisisdano: Used in the breadcrumb nav

docs/src/components/base/type-scale/serif/serif.html
```diff
@@ -0,0 +1,59 @@
...
+</div>
+
+<div class="pattern">
+  <div class="pattern-example font-serif lh-single serif-s5">Cumulonimbus and cirrus</div>
+  <div class="pattern-title sans-s5 fw700">17px</div>
+  <div class="pattern-code mono-s3 fw300">.serif-s5</div>
+</div>
+
+<div class="pattern">
+  <div class="pattern-example font-serif lh-single serif-s6">Cumulonimbus and cirrus</div>
```
msecret, line 32: Where are all the `pattern` classes coming from?

docs/src/components/base/layout/README.md
```diff
@@ -0,0 +1,43 @@
...
+  <div class="grid-width-8"></div>
+  <div class="grid-width-4"></div>
+</div>
+```
+
+## Flex layout
+The flex layout has no max width. It is composed of parent `row`s and child `col`s. Each row of columns needs to be enclosed in a `row` and **columns don't have to add up to anything**. If you're using a pure flex `-f{#]` grid, the toal number of columns is simply the sum of the `-f{#}`s. (you can mix and match with `col-fit` and `col-fill` as well, _or_ have underfilled rows using fixed or percentage width utilities). Columns may include nested rows and columns as in the following:
+
+```html
+<div class="row">
```
msecret, line 26: I'm still thinking about this but am a little unsure of it. I mentioned I didn't like `f` from a naming perspective. I was thinking: what if we just went with `col`? Then if down the road we ditch flexbox for the newer and greater thing, we don't have to rename our classes?

docs/src/components/base/borders/borders--2px.html
```diff
@@ -0,0 +1,7 @@
...
+    <div class="swatch h20 mb8">
+      <div class="example swatch-color b2px b-{{ color.colorName }}"></div>
```
msecret, line 3: I don't think `b` as a name for bordered is understandable enough. Also, where are those classes defined?

src/css/override_vars.scss
thisisdano: Seems there's a compromise between succinctness and readability — in the world of utility classes, I tend to want to err in the direction of succinctness.

There are [n] scales in the utility classes:

grid scale: [#]
type scale: s[#]
flex scale: f[#]
pixel scale: [#]px
percentage scale: [#]ct 

To me, this feels right.

However, if you want the actual class name in the standard scss to be more like `.col-flex-1` — that seems ok.

Actually, this `$u-flexscale` var is only used in one place — `utility-classes.scss:420` and that could be rewritten...

docs/src/components/base/type-scale/serif/serif.html
thisisdano: `main.css` in the PL — they're just helper classes for PL display

docs/src/components/base/colors/colors.html
```diff
@@ -0,0 +1,8 @@
+{% for color in colors %}
+  <div class="swatch h20 mb8">
+    <div class="example swatch-color bg-{{ color.colorName }}"></div>
```
msecret, line 3: So I feel like a lot of the code in `base` is using a very different perspective of writing CSS then the rest of c-style, namely a Atomic CSS way. While I think there's room for Atomic CSS in certain circumstances in cg-style, I think it should be limited in base styling, potentially like this. I'm also not a fan of using `div`s for marking up text, as I'd prefer headings and ps. 

In terms of Atomic CSS; I think it's use case works best for one of a kind or one off layout rules. Such as I have a component that has a good set of rules and just this one time I need it to move over to the right, that's a good flexbox rule.

So is there a reason we're using these types of rules here? Could we make this work with `h`s and `p`s and such?

docs/src/components/base/borders/borders--2px.html
thisisdano: Again — I think this is acceptable in the world of utility classes. I'm not sure how much we will be using them in production — they're mostly a prototyping tool for the PL at this point, and since you have to type them frequently, they're short and necessarily have some learning curve...

IDK — I look to something like basscss or tachyons, or solid for taxonomy guidance, and they're all pretty succinct.

docs/src/components/base/colors/colors.html
thisisdano: I don't see that it makes much of a difference how we display the color rules in the PL as these swatches don't have a production equivalent. 

I use divs because I mostly don't care about the semantic value of the structure when I'm making the prototype — I assume that others have a stronger opinion and will modify when necessary. Divs are just a simple way of getting the material down on the page.

docs/src/components/base/layout/README.md
thisisdano: I could accept something like `col-1` — I like the `f` prefix to create distinction with my other scales, but in this case there might not be any realistic ambiguity.

docs/src/components/base/borders/borders--2px.html
thisisdano: Well, this I'm doing to properly compose fractal variants. If you name it without the `--` it won't appear as a variant

src/css/components/grid.scss
thisisdano: This is all for the the utility classes — which are all meant to be overrides. I imagine in most cases you might not use these tools for production components, but if you did, you'd want it to override. In a world with 100% utility classes, the important would be redundant, tho.

src/css/components/trigger.scss
thisisdano: That is what it is — a dropdown trigger.

thisisdano:
In general, my breadcrumb-style PR (https://github.com/18F/cg-style/pull/228) builds off of this, so it might make sense to make any agreed upon changes to that branch/PR. Otherwise I'll find myself in a potential sea of merge conflicts....

docs/src/components/base/borders/borders--2px.html
thisisdano: all utility classes are defined in `utility-classes.scss`

docs/src/components/base/colors/colors.html
thisisdano: Base is just where I'd expect to use this kind of classing since the items in this section are below the pattern level and need no special classes developed for their presentation — like, for instance, `.pattern-swatch` or some other thing that's only used for PL content, never in the production code.

Pretty much agree wrt utility classes in production (at least on this project) — I think there's a good case for them to be used more broadly but that ship has sailed in this instance. The other good time to use utility classes is for prototyping — and this is mostly how I've used them here — to develop new patterns, or to make pattern suggestions without touching or modifying any of the styles used by live code.

src/css/components/card.scss
thisisdano: Yeah, that sounds fine

src/css/components/icon.scss
thisisdano: This makes sense

src/css/override_vars.scss
msecret: I'd prefer readability, even in utility classes, as that's the most common problem I've seen when new developers take over a project.

line47: I'd agree with readability, but also feel there is a good middle ground.

thisisdano: Aight, so here's my additional 2 cents: I think there's a very high value to succinctness in utility classes (for scss variables, less so). For a new dev — or any dev — getting familiar with these classes there will be a learning curve and necessitate documentation. Nothing will be 100% obvious without reading documentation first.

So, I'd propose leaving succinct utility classes, but making more verbose scss variables and non-utility classes. AND putting more effort into clearly documenting the utility classes we have — which are now legion.

src/css/override_vars.scss
thisisdano: Agreed in this instance.

adborden:
Oops, we're accidentally adding back `documentation/.jekyll-metadata`. You can just remove your `documentation/` dir since it's no longer needed.

thisisdano:
OK, removed /documentation folder

docs/src/main.css
adborden: ah, right, this is for the [grid aka layout preview](http://localhost:3000/components/detail/layout) so that the text shows on a light background. But looks like the other classes are affecting the color now, so we can remove the `background-color` and `color` properties here.

adborden:
If we're using the component status (draft, ready, wip), what do those mean in our workflow? Can `wip` be used on the dashboard?

thisisdano:
^^^ TBD — that's something we haven't discussed

thisisdano:
dunno difference between WIP and draft — draft is, like, a completed WIP?

The way I see it is that WIP/Draft is for patterns moving from design to production — i.e. those that are prototypes that need to be converted to react templates and perhaps given permanent classes. Once that's done, the pattern could be updated in the PL to reflect the final structure/classes.

thisisdano:
Kinda why I initially pushed to have a `Prototypes` section — so there's no confusion about what's a production component

src/css/components/icon.scss
thisisdano: I'd like to punt this until we have more UI SVGs — I don't know yet how they would best be classed

line47:
I have a few thoughts about statuses that I think make sense for our workflow.

- WIP - Should be a prototype of sorts still in flux and being created.
- Draft - Completed prototype that needs some review.
- Ready - Complete and reviewed by another team member. Can be implemented.

docs/src/components/base/layout/README.md
thisisdano: That said — I don't know the utility of planning too far ahead for functionality that doesn't exist. The generalism model is somewhat counter to the utility class model. I like that this utility class is specifically tied to its flex properties. If a new bit of layout functionality appeared, we'd probably want a new utility class for it.

docs/src/components/base/spacing-scale/README.md
```diff
@@ -0,0 +1,78 @@
...
+$grid-2:             2 * base       16px
+$grid-3:             3 * base       24px
+$grid-4:             4 * base       32px
+$grid-5:             5 * base       40px
+$grid-6:             6 * base       48px
+$grid-7:             7 * base       56px
+$grid-8:             8 * base       64px
+$grid-9:             9 * base       72px
+$grid-10:            10 * base       80px
+```
```
adborden, line 23: This might present better in a markdown table (I haven't seen how our fractal theme presents them).
```
| Class | Scale | Computed |
| -------- | ------- | -------------- |
| `$grid-05` | 0.5 * base | 4px |
```

| Class | Scale | Computed |
| -------- | ------- | -------------- |
| `$grid-05` | 0.5 * base | 4px |

thisisdano: Agree — I want to totally update a lot of this documentation and this is probably how i'd want to do it.

thisisdano:
@line47 But do we then need a status for an complete/implemented pattern?

thisisdano:
I'm finding it hard to keep up with code review comments, and losing comments I KNOW were in the PR. Does it feel buggy to others? I'd love for it to work for us, but I fear there are conversations that are getting lost...

thisisdano:
What needs to be done to the PR at this point to get it merged? 
What key issues need to be resolved now and which can be punted to a subsequent PR?

(I'd like to pull these key conversations out of the review comments for the reasons noted above.)

docs/src/components/features/card-home-contents/card-home-contents.config.yml
```diff
@@ -0,0 +1,29 @@
...
+        - space:
+          name: 'foo/bar'
+      apps:
+        - app:
+          name: 'foo/foo'
+        - app:
+          name: 'foo/bar'
+    - org:
+      name: 'epa-journeyman'
+      isOpen: true
```
adborden, line 20: Hmm.. I'm not seeing this one rendered in the open state.

src/css/base/utility-classes.scss
```diff
@@ -0,0 +1,1767 @@
...
+  Modifiers:
+    ---
+
+  Values:
+    {$color}
+
+*/
+
+$color-s: 'blue', 'brightmint', 'cream', 'gold', 'green', 'red', 'trueblack', 'textblack', 'darkgray', 'mediumgray', 'mediumgray-trans', 'lightgray', 'lightestgray', 'white';
+$colorvar-s: $color-blue, $color-brightmint, $color-cream, $color-gold, $color-green, $color-red, $color-trueblack, $color-textblack, $color-darkgray, $color-mediumgray, $color-mediumgray-trans, $color-lightgray, $color-lightestgray, $color-white;
```
adborden, line 32: For stuff like this, using a sass map might be better because the label and values line up. sass has functions for iterating over maps or fetching a single value from the map.

src/css/base/utility-classes.scss
```diff
@@ -0,0 +1,1767 @@
...
+.font-sans { font-family: $font-sans !important; }
+
+%font-serif { font-family: $font-serif; }
+.font-serif { font-family: $font-serif !important; }
+
+%font-mono { font-family: $font-mono; }
+.font-mono { font-family: $font-mono !important; }
+
+%font0 { font-family: $font-sans; }
+.font0 { font-family: $font-sans !important; }
```
adborden: What do the `0`'s mean in class names? I don't really see the pattern between `bg0`, `sans0`,  and `font0`.

src/css/base/utility-classes.scss
```diff
@@ -0,0 +1,1767 @@
...
+*/
+
+%measure-reading { max-width: $text-max-width; }
+.measure-reading { max-width: $text-max-width !important; }
+
+%measure-narrow { max-width: $text-narrow-width; }
+.measure-narrow { max-width: $text-narrow-width !important; }
+
+%measure0 { max-width: none; }
+.measure0 { max-width: none !important; }
```
adborden: We definitely want to add some docs about the vocabulary :) Either mention that we're using terms from print/copy industry and link to a print glossary, or include the basic terms... I would prefer to include the terms in our docs.

src/css/base/utility-classes.scss
```diff
@@ -0,0 +1,1767 @@
...
+
+.clearfix:after,
+.clearfix:before {
+  content: ' ' !important;
+  display: table !important;
+}
+
+%clearfix:after { clear: both; }
+
+.clearfix:after { clear: both !important; }
```
adborden: Do these already exist in one of our libraries? Maybe wds?

src/css/base/utility-classes.scss
```diff
@@ -0,0 +1,1767 @@
+@import '../core.scss';
+
+// base utility classes, primarily for documentation and prototyping
+
+$u-percentages: '5', '10', '20', '25', '30', '33', '40', '50', '60', '66', '70', '75', '80', '90', '100';
+$u-percentage-suffix: 'ct';
+$u-percentages-s: '5%', '10%', '20%', '25%', '30%', '33.3333333333%', '40%', '50%', '60%', '66.6666666666', '70%', '75%', '80%', '90%', '100%';
+$u-pixels: '1px', '2px';
+$u-gridscale: '05', '1', '105', '2', '3', '4', '5', '6', '7', '8', '9', '10', '15', '20', '25', '30', '40', '50';
```
adborden, line 9: Don't some of these clash? For example I think `.w50` is defined for both a percentage width and a grid width.

thisisdano:
@msecret @adborden @line47 ^^^

src/css/reset.scss
```diff
@@ -0,0 +1,117 @@
+/* http://meyerweb.com/eric/tools/css/reset/
```
adborden: Looks like we can get this from [npm](https://www.npmjs.com/package/reset-css), so we can avoid including this file.

adborden:
>I'm finding it hard to keep up with code review comments, and losing comments I KNOW were in the PR. Does it feel buggy to others? I'd love for it to work for us, but I fear there are conversations that are getting lost...

Github will hide the comments when the line the comment referred to was changed. The assumption is that if you changed the line, you addressed the comment. They are still there, in order, you might just need to expand the diff.
peek 2016-12-22 11-41

adborden:
@thisisdano we can use checklists to itemize specific changes so that everyone is on the same page. Right now I am working on two lists. 

- [ ] list of issues that need addressing before this PR is merged
- [ ] list of issues that should be addressed in a new PR

adborden:
Oops, I forgot to speak to your other point. I would prefer to keep the conversation in this PR. Otherwise we're asking to let things slip through the cracks. But it would help if we use the checklists and are explicit about what changes we are looking for. That way Dan can check things off the list and the reviewers can look back at the checklist and confirm that each change has been made before giving the :+1:

src/css/components/trigger.scss
thisisdano: Styleguide updated.

line47:
@thisisdano is say the `Ready` status should be used as completed patterns that can be implemented in the code.

src/css/base/utility-classes.scss
msecret: I think it does

thisisdano:
What about this kind of status system:

**Prototype** (testing, farting around, maybe wouldn't be implemented)
**Review-ready** (prototype ready for implementation discussion)
**In Progress** (Making changes to get ready for implementation)
**Ready** (Ready to implement)
**Active** (Active component in the production app)

Overkill?

line47:
I think that's a good pass on status. Speaking over overkill...? 
Would we want to have a **deprecated** status?

thisisdano:
**Deprecated** seems like a good addition to me.

so 

**Prototype** (testing, farting around, maybe wouldn't be implemented)
**Review-ready** (prototype ready for implementation discussion)
**In Progress** (Making changes to get ready for implementation)
**Ready** (Ready to implement)
**Active** (Active component in the production app)
**Deprecated** (no longer active)

docs/src/components/component/button/button.config.yml
adborden: fwiw, I like what you do with the swatches, there's the component itself, and then there's some helper text so you can see the name, as well as a realistic example.

msecret:
For statuses:

I don't understand the difference between `Active` and `Ready`. I'd prefer to just `Ready` in that this component is completely ready to use. I'm also a little confused about `Review-ready` and `In progress`. Those almost sound more like PR statuses rather the component statuses. I think `Review-ready` also insinuates a review process, which we currently don't have. I like the idea of separating between a prototype, a component that is implemented correctly in code but needs design review /accessibility review, a component that has those reviews but hasn't been used extensively so might change a lot and then a mature component. Maybe the last Mature one isn't needed actually.

- Prototype - Code implementation is in prototype phase, incomplete.
- Draft - Code implementation is a little more complete, design/accessibility reviews have not been done, documentation incomplete
- Beta - When some design/accessibility reviews have been done and some documentation is complete but not all of it is. The component is ready to use at this point technically.
- Ready - Accessibility review finished, usability review finished, documentation complete.
- Deprecated

thisisdano:
Honestly, as long as we just pick something and try to work with it, I don't really care enough to argue. It's all for stuff we won't be using anyway. 

Love to follow up on getting this PR merged, it's been ffe and this is holding up future work and being out-of-sync is making things confusing for me and possibly others.

> What needs to be done to the PR at this point to get it merged?
> What key issues need to be resolved now and which can be punted to a subsequent PR?

docs/src/components/components/breadcrumb-nav/bc-app-view/bc-app-view--appOpen.html
```diff
@@ -0,0 +1,251 @@
+{# breadcrumbNav#}
+<nav class="bg-white p4 mn2">
```
msecret, line 2: So the breadcrumb is still in prototype mode?

msecret:
Issues that I think should be fixed before merge:

- [x] There's still naming inconsistencies from component file name to component classes to pattern library component name. There's probably more then this even:

  - [x] `card-home-contents` to just `card`. I made a comment herehttps://github.com/18F/cg-style/pull/225/files#r93805217. **[done]**
  - [x] Button Trigger should be renamed to just Trigger **[done]**
  - [x] Overview Row to Overview **[done]**

- [x] Theres still a lot of components that are using atomic classes in their PL markup. I think these should either get a "Prototype" status or have their atomic classes removed. I think all the features and pages also have this problem:

  - [x] Panel **[already marked as WIP]**
  - [x] Overview **[already marked as WIP]**
  - [x] Breadcrumb **[already marked as WIP]**
  - [x] Button Trigger (should be renamed to just Trigger) **[already marked as WIP]**
  - [x] Count Status **[already marked as WIP]**

- [x] I think the changes made in `package.json` should be removed, as they were put there before @adborden fixed those issues. **[replaced with `package.json` from master]**

- [x] The `components/grid.scss` file is in `components` but is supposed to be a utility? Should it be moved? We should have `!important` rules in our components. **[left here because the other grid was here, but happy to move]**

- [x] Is the reset file required right now? We already have a normalize css in use, and we probably shouldn't have both. Can we just remove the reset file for the time being until we decide which one we want? **[removed]**

- [ ] Ensure the utility file isn't included on production. **[I don't know how to do this myself — can this wait for a separate issue?]**

Not for now but something we have to figure out really soon:

- [ ] Figure out what to do about the utility file:

The utility file is 1k lines of code, which even if it won't be used in production, is a lot of code to maintain. On top of that, the naming convention isn't 100% and even has some conflicts. Rather then maintain this file, how do we all feel about adding in an outside atomic CSS framework that will only be includes in the pattern library CSS? This means we don't have to write and maintain documentation for our own file and can rely on their docs and such. Thoughts? **[I developed this system to accommodate the level of detail our patterns require — and to keep the styles consistent with our Sketch system. I don't think another acss system will be able to do just what this one does. Are there gaps? Sure. Inconsistencies in the naming convention? Maybe a couple, but vv few tbh — no worse than any other system, certainly no worse than that of the cg styles in general. Conflicts? IDK, I haven't seen any. I _would_ like to move the acss framework out of cg and into its own non-cg repo for use on non-cg projects, but that's a longer process. Also, happy to include the utility classes only in the PL.]**

src/css/base/typography.scss
```diff
@@ -3,9 +3,14 @@
 
 body {
   color: $color-base;
+  font-feature-settings: 'kern';
```
msecret: What are all these new settings doing? Some of them do not seem cross-browser. What is there browser support?

src/css/components/card.scss
```diff
@@ -0,0 +1,27 @@
...
+
+.card {
+  background-color: $color-white;
+  border: 1px solid $color-textblack;
+  border-radius: $rounded;
+  margin-bottom: $grid-2;
+  padding: 0;
+}
+
+.card-home-contents-title {
```
msecret: These classnames are too deeply nested. Why can't we remove the `home` and `contents` sections, especially considering these cards are only being used in one place.

adborden:
@thisisdano I updated the PR description with the lists from the comments within the PR. These things are obviously up for discussion, but most of them we seemed to have agreement on in the comments.

@msecret can you take a look and make sure we feel good about the now and later stuff?

msecret:
I additionally had some thigns to add in comment, https://github.com/18F/cg-style/pull/225#issuecomment-269049790. I can add so everything is in same list though?

adborden:
@msecret yeah, let's keep it all at the top if that works for everyone. I already added the items from your comment.

adborden:
@thisisdano for the layout pattern, my eyes are drawn toward the gutter, rather than the columns themselves. Maybe if we put the pattern on a grey/darker background it would be easier to see the columns?

http://localhost:3000/components/detail/layout
screenshot from 2016-12-23 14-24-22

docs/src/components/components/breadcrumb-nav/bc-app-view/bc-app-view--appOpen.html
thisisdano: yes, yes, 1000x yes

src/css/base/typography.scss
thisisdano: 1. I rely on autoprefixer for x.browser, but http://caniuse.com/#feat=font-feature
2. http://blog.typekit.com/2014/02/05/kerning-on-the-web/
3. It's progressive enhancement, so if it fails on old browsers, it's not critical
4. looks like it should actually be `font-feature-settings: 'kern' 1;` so I'll make that change.

src/css/components/card.scss
thisisdano: tbh, i don't care. I can change them to be less specific, but i don't even know if these styles are in use at all. this is one of those things where I build something to try and show how it could work (as a prototype, essentially) and I don't think anything made it into production.

So sure, I'll change it.

src/css/base/utility-classes.scss
thisisdano: IDK what this is — it looks a like the result of a mis-applied find/replace

thisisdano: previously, this was `.font--` which was the shorthand for the default value.

src/css/base/utility-classes.scss
thisisdano: i will look up how to to do this. tho does it need to be a part of this PR — why isn't this a nice to have that we can add later?? :(

src/css/reset.scss
thisisdano: gonna chuck this whole file since I never really figured out hoe to use it in the way I intended.

src/css/base/utility-classes.scss
thisisdano: removed

src/css/base/utility-classes.scss
thisisdano: What would the non-print equivalent to this be? `line-length`?

thisisdano: changed to line-length (`ll`)

src/css/components/card.scss
thisisdano: FYI, I don't agree that we should rename the feature 'card' — the feature is still this specific implementation/variant of the card.

src/css/base/utility-classes.scss
thisisdano: I don't think this is true. `w50` should be 50xgrid width. 50% width would be `w50ct`.

thisisdano:
Does all the requested stuff NEED to be in before the PR can be merged?

The strict scrutiny on this PR seems to go above and beyond any reasonable MVP requirements to a level of fussiness that feels obstructionist and perhaps unintentionally rude. I agree with a good number of the comments, and could see them as reasonable future issues and reasonable tech debt. I've reasonably responded to almost every issue here, either with a change or with some form of reasoning. Almost everything remaining falls in the category of nice-to-have.

I have tried to be quick and reasonable in my responses to this PR, yet it has sat in the queue for over 10 days, and I'm sure it will stay there longer. Every new review is a new opportunity for new requests and discussions that go unresolved. It's far from a constructive PR process. Every day this sits here is another day that reasonable work — critical work, even — based off this PR sits blocked and undone. I mean, 106 comments and growing. That's unusual.

I strongly suggest we move ahead to merge this PR and convert any lingering issues here into proper issues and deal with them in a more structured, practical way — it feels more in the spirit of iterative progress than this protracted quagmire of endless tar pit games.

If there are real reasons not to proceed in this fashion, let's hear them. I will continue to address existing PR issues in the interim.

thisisdano:
OK, looked through the Sass maps functionality and determined that it's beyond my present abilities to a) determine if and how this would make our code better; and b) implement it in a way that would make our code better. There's a fair amount of complexity that could make for more efficient code, but it would take real work to determine how this efficiency could best be used to make the code simpler to maintain. It may just be more complicated.

So, I file this under nice-to-have and punt to the future.

package.json
```diff
@@ -52,7 +49,6 @@
   },
   "homepage": "https://github.com/18F/cg-style#readme",
   "dependencies": {
-    "@frctl/nunjucks": "^1.0.2",
```
adborden: I think we still need this line, but we should move it to `devDependencies`.

thisisdano: what's the difference between these two sections?

thisisdano: Fixed

msecret (approved):

PR #226, opened 2016-12-16
Fixes layout of old and new UI

msecret:
These fixes allow for all panel layouts, in the old table code and new panels to be consistent and closer to the design.

### DIFF
--- a/src/css/components/panel.scss
+++ b/src/css/components/panel.scss
@@ -7,7 +7,7 @@ $panel-padding: $grid-3;
 
 .panel {
   @include clearfix;
-  min-height: 10rem;
+  min-height: 6rem;
   padding: .75rem 0 0;
 
   & + .panel {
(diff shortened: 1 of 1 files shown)


src/css/components/panel.scss
```diff
@@ -28,11 +33,11 @@ h4 + .panel {
 
 .panel-title {
   font-size: 1.3rem;
-  margin-bottom: 2rem;
+  margin-bottom: 1rem;
 }
 
 .panel-group {
-  margin-bottom: .65rem;
+  margin-bottom: 0.65rem;
```
adborden, line 40: Can we turn this into a linting rule?

adborden (approved):

src/css/components/panel.scss
line47: I think we should if we are keeping the leading zero. 
https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#leadingzero

PR #229, opened 2016-12-20
Adding style for usage and allocation and pattern for status pill

line47:
Added a new SCSS file for the status pill.
Updated the activity log and status SCSS to use the pattern for status pill.

### DIFF
--- a/src/css/components/activity_log.scss
+++ b/src/css/components/activity_log.scss
@@ -1,3 +1,4 @@
+
 @import '../core.scss';
 @import '../override_vars.scss';
 @import '../vars.scss';
@@ -35,32 +36,27 @@
   cursor: pointer;
   display: flex;
   flex-wrap: wrap;
-  margin-bottom: .5rem;
-  margin-top: .5rem;
-  padding-bottom: .5rem;
-  padding-left: 1rem;
+  margin: 0;
+  padding: .5rem 1rem;
   position: relative;
   width: 100%;
 
   &:first-child {
     border-top: 1px solid $color-lightgray;
-    padding-top: .5rem;
+
   }
 }
 
-.activity_log-item-error,
-.activity_log-item-success,
-.activity_log-item-warning {
-  .activity_log-item_text::before {
-    border-radius: 3px;
-    border-style: solid;
-    border-width: 2px;
-    content: '';
-    height: 100%;
-    left: -1rem;
-    position: absolute;
-    top: 0;
+.activity_log-item-error::before,
+.activity_log-item-success::before,
+.activity_log-item-warning::before{
+
+      @extend .vertical-pill;
   }
+
+  .activity_log-item-error,
+  .activity_log-item-success,
+  .activity_log-item-warning {
 
   .activity_log-item_text {
     font-family: $font-sans;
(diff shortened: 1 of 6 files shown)


src/css/components/activity_log.scss
```diff
@@ -35,33 +36,28 @@
...
-  padding-left: 1rem;
+  margin: 0;
+  padding: .5rem 1rem;
   position: relative;
   width: 100%;
 
   &:first-child {
     border-top: 1px solid $color-lightgray;
-    padding-top: .5rem;
+
```
msecret: This line can be removed.

src/css/components/status_pill.scss
```diff
@@ -0,0 +1,13 @@
+
+@import '../override_vars.scss';
+@import '../core.scss';
+
+.vertical-pill {
```
msecret: The name of the CSS class and the file name should probably match. Can we call `vertical-pill` `status-pill`, or other way around?

msecret (commented):
This code looks great, such a big improvement to remove all that duplication. Do you think it makes sense to also extend the border-color changes based on status, or is that not needed?

You also have some linting errors in `activity_log` scss file.

src/css/components/status_pill.scss
line47: Sure thing. I opted for `status-pill` as it's not tied to presentation in any way.

line47:
I think we can explore extending the border color changes. 
I'm open to suggestions on that as well. 

I'm thinking it would make sense to incorporate `status-pill.SCSS` into `status.SCSS` as well and do all of the status color changing within `status.SCSS`.

src/css/components/panel.scss
```diff
@@ -22,6 +22,17 @@ $panel-padding: $grid-3;
   .tableWrapper:last-child {
     margin-bottom: 2rem;
   }
+  .subtext {
```
msecret: Why is this part of the status pill PR?

line47: This is part of the usage and allocation panel updates as well.

msecret: K. So we have to try and stop adding things under `.panel`. Is `subtext` an existing class? What's it's purpose? Rather the adding different styling for the `h2`, it might make more sense to add a `sans-s3` class to the `h2` in the html.

msecret (changes requested):

src/css/components/panel.scss
line47: `subtext` is for the "Instance applies to memory and disk" text up in the number of instances column. I don't think this text should be an `<H2>`. Maybe it should be styled as `stat-info`?

line47: I removed the `subtext` class

msecret (approved):

src/css/components/stat.scss
```diff
@@ -13,77 +13,78 @@ $input-font-size: $sans-s2;
...
   min-height: 2.5rem; // Leave some space for input element when editing a stat
+  position: relative;
   width: 100%;
 
   .error_message {
     font-size: $input-font-size;
   }
 }
 
+h5 + .stat {
```
adborden, line 24: Is this rule still used? I think we're getting rid of all the `h5`s in this PR https://github.com/18F/cg-dashboard/pull/855

src/css/components/panel.scss
```diff
@@ -22,6 +22,13 @@ $panel-padding: $grid-3;
   .tableWrapper:last-child {
     margin-bottom: 2rem;
   }
+
+  h2 {
```
adborden, line 26: I think we should be avoiding these rules and use class modifiers instead.

PR #231, opened 2016-12-22
Changes to accommodate org overview page

msecret:
Adds a subtext style, I think @line47 needed this as well. I needed another panel header style for the smaller, bold type for when it's a table. I want to abstract this away at some point but wanted to be sure we have more design ready before doing so. I simplified the panel-row code so certain layout things happen at just two nested levels of panel-rows. This reduces complexity and ensures the space overview and org overview don't need a fake extra panel row. It required adding some extra rules for the overview page, which is related to the boxed panel row.

The panel code still has to be refactored, I wanted to do so when the big PL PR is done, and when we've seen all design. From there, we should look at all the different parts we're using and abstract things away.

### DIFF
--- a/src/css/components/count_status.scss
+++ b/src/css/components/count_status.scss
@@ -5,7 +5,7 @@
 
 .count_status {
   display: inline-block;
-  height: 0.6rem;
+  height: 1.1rem;
   margin-left: $grid-3;
   position: relative;
   width: 7.5rem;
--- a/src/css/components/panel.scss
+++ b/src/css/components/panel.scss
@@ -182,6 +182,12 @@ h4 + .panel {
   min-width: 25%;
 }
 
+// TODO rename `-less` to something else
+.panel-column-shrink {
+  flex-grow: 0;
+  margin-right: 2rem;
+}
+
 .panel-new {
   background-color: $color-white;
   margin-top: $grid-4;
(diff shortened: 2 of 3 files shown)


adborden:
I think https://github.com/18F/cg-dashboard/pull/855 depends on this.

src/css/components/count_status.scss
```diff
@@ -5,7 +5,7 @@
 
 .count_status {
   display: inline-block;
-  height: 0.6rem;
+  height: 1.1rem;
```
adborden, line 8: What's with the non-round number? Could this be based on our grid-sizing hierarchy?

adborden (approved):

src/css/components/panel.scss
```diff
@@ -59,6 +59,11 @@ h4 + .panel {
   }
 }
 
+.panel-actions {
+  display: block;
+  margin-top: 1rem;
```
line47: Wouldn't we want to use `$grid-` here for the margin?

src/css/components/panel.scss
```diff
@@ -182,6 +202,12 @@ h4 + .panel {
   min-width: 25%;
 }
 
+// TODO rename `-less` to something else
+.panel-column-shrink {
+  flex-grow: 0;
+  margin-right: 2rem;
```
line47: Wouldn't we want to use `$grid-` for the margin?

PR #240, opened 2016-12-27
new styles for homepage columns

thisisdano:
Added a couple styles to color the borders on the new homepage

### DIFF
--- a/src/css/components/three_up.scss
+++ b/src/css/components/three_up.scss
@@ -8,6 +8,7 @@
 .three_up-part {
   margin-top: 4rem;
   padding-top: $grid-3;
+  border-top: 2px solid;
 
   @include media($medium-screen) {
     @include span-columns(4);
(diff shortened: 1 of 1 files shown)


thisisdano:
Fixes change requests made here: https://github.com/18F/cg-site/pull/661

src/css/components/three_up.scss
```diff
@@ -22,19 +23,15 @@
     margin-bottom: 1rem;
   }
 
-  &.bt2px {
-    border-top: 2px solid;
-  }
-
-  &.b-red {
+  &.part-3 {
```
msecret: The convention we've been using in our components is that elements of a component (like this part) are named with the component name so:

`.three_up-part-1`, etc.

Is there a reason you want to do something different here?

thisisdano: nah, I'll just rename

adborden (approved):

PR #242, opened 2016-12-28
First pass at utility docs and overview fix

thisisdano:
This adds a  first pass at utility class docs. There's definitely more that can/should be done, but this lays the foundation. It also fixes that bug w/ `features/panel-overview-org`

screen shot 2016-12-27 at 5 08 07 pm


adborden (approved):
Great, this is very helpful!

PR #243, opened 2016-12-28
New styles for PL docs

thisisdano:
Style changes for docs readability

- - -

screen shot 2016-12-28 at 8 00 57 am

### DIFF
--- a/docs/src/docs/01-index.md
+++ b/docs/src/docs/01-index.md
@@ -1,12 +1,7 @@
 ---
-title: cloud.gov design system
+title: Overview
 status: draft
 ---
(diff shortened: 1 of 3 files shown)


docs/src/docs/01-index.md
```diff
@@ -1,18 +1,13 @@
 ---
-title: cloud.gov design system
+title: Overview
```
adborden, line 2: Since this is the first thin you see, it might be good to have this be something like "cloud.gov Pattern Library" and Overview as the first header.

adborden (approved):

PR #247, opened 2016-12-29
A few changes to status display, and a first pass at a more expressive status bar prototype

thisisdano:
This changes the display of status dots, modifies status on some items, and includes a prototype of a count status components with more expressive count status.

- - -

screen shot 2016-12-29 at 7 47 33 am


docs/src/components/components/count-status-grid/count-status-grid.config.yml
```diff
@@ -0,0 +1,23 @@
+title: count-status-grid
+status: prototype
+context:
+  entityKind: space
+  status: alt
+  entityCount: '1'
+variants:
+  - name: default
```
adborden, line 8: btw, when there isn't really a default, you can just specify `default: app` to name the app [variant as the default](http://fractal.build/guide/components/variants#the-default-variant)

adborden (approved):

PR #251, opened 2016-12-29
Align top nav and disclaimer with dashboard width [IGNORE FOR NOW]

thisisdano:
**Ignore in favor of dropping the sidebar**

This PR adjusts the padding and max-widths of the disclaimer and the top nav of the dashboard to align their left and right edges with the outside edges of the dashboard content. It adjusts the `$medium-screen` responsive width to better determine when the `disclaimer-stage` text should show. The end result is a cleaner visual presentation of the top of the page.

This affects only dashboard with a kludgey style hook: `.js-app`. I used this hook only because it was an existing class that is unique to the dashboard. We could apply a better hook, perhaps at the `body` level, but I didn't want to touch `cg-dashboard`. 

Can someone help me determine where to put this style hook and what it should be?

- - -

screen shot 2016-12-29 at 10 45 59 am

### DIFF
--- a/src/css/components/header.scss
+++ b/src/css/components/header.scss
@@ -16,8 +16,7 @@
   padding-bottom: $grid-3;
 }
 
-// kludge for targeting dashboard
-.js-app {
+.header-wide {
   .header-wrap {
     max-width: 1224px;
     padding-left: $grid-4;
--- a/src/css/components/header.scss
+++ b/src/css/components/header.scss
@@ -16,6 +16,15 @@
   padding-bottom: $grid-3;
 }
 
+// kludge for targeting dashboard
+.js-app {
+  .header-wrap {
+    max-width: 1224px;
+    padding-left: $grid-4;
+    padding-right: $grid-4;
+  }
+}
+
 .header-title {
   float: left;
   margin-bottom: .6rem;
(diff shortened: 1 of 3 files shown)


src/css/components/header.scss
```diff
@@ -16,6 +16,15 @@
   padding-bottom: $grid-3;
 }
 
+// kludge for targeting dashboard
+.js-app {
```
msecret: Rather then add more kludge, why don't we add another header class, like `.header-lg` or `header-full`.

src/css/override_components.scss
```diff
@@ -45,6 +48,22 @@
   }
 }
 
+// kludge for targeting dashboard
+.js-app {
```
msecret: Same as above, maybe `disclaimer-lg`, or wrap the disclaimer in the `header-lg` or whatever class.

msecret (changes requested):

src/css/components/header.scss
thisisdano: Think I'll do this, wrap in `header-wide`

src/css/override_components.scss
thisisdano: Think I'll do this, wrap in `disclaimer-wide`

thisisdano:
@msecret fixed the class hooks per your request. Now paired with https://github.com/18F/cg-dashboard/pull/884

PR #258, opened 2016-12-30
Updates to pattern library for status and moving components

line47:
Moved some more complex items under features like card and panel. Also
updated and added some basic read for actions.


docs/src/components/components/action/README.md
```diff
@@ -0,0 +1,11 @@
+
+# Action
+
+The **action** component is something to use when you want the user to take an
+action. Buttons should be used for changing the state of a page or transforming
+data. Links should be used for navigation.
+
+## Use
+
+To use the **action** component, add the class `action` to a button or add the
```
thisisdano: Worth being specific that this needs to go on a `<button>` element — this is _almost_ specific enough! :)

docs/src/main.css
```diff
@@ -378,7 +379,6 @@ body._hasBackground {
 .Prose code pre {
   font-family: 'Roboto Mono';
   background-color: transparent;
-  width: 100%;
   display: block;
 }
```
thisisdano, line 384: These last changes _probably_ screw up some of my utility class docs (I'd made these changes to make the code blocks in that doc look right...)

thisisdano (commented):
Couple quick things...

docs/src/main.css
line47: The 100% width shouldn't be needed as display:block does that for us. Looks good on my local. Can you verify locally?

docs/src/components/components/action/README.md
line47: Good call, I'll update.

docs/src/main.css
thisisdano: If it looks good to you, that's good enough for me.

thisisdano (approved):

PR #261, opened 2017-01-03
Simple breadcrumbs

adborden:
This should be merged after
- [ ] https://github.com/18F/cg-style/pull/260 page header

screenshot from 2017-01-02 16-44-27

### DIFF
--- a/docs/src/components/components/breadcrumbs/README.md
+++ b/docs/src/components/components/breadcrumbs/README.md
@@ -1,0 +1,3 @@
+# Breadcrumbs
+
+The current page should not be represented in the breadcrumb.
--- a/docs/src/components/components/breadcrumbs/breadcrumbs.config.yml
+++ b/docs/src/components/components/breadcrumbs/breadcrumbs.config.yml
@@ -1,0 +1,12 @@
+title: Breadcrumbs
+status: beta
+context:
+  breadcrumbs:
+  - text: ''
+    icon: home
+  - text: my-org
+    icon: org
+  - text: my-dev-space
+    icon: space
+  - text: my-app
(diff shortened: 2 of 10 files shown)


adborden:
There's a few things that stick out to me,
- Is there a class to use to take care of the margins? It seems strange that panels handle the left/right padding. I'm using the grid here, but it doesn't account for the gutters. I would use panel, but it adds the white background. I feel like I had addressed something long ago between `.content` and `.grid` to handle this, but I can't remember what that is.
- I couldn't decide on the link color for the bread crumbs. we're using blue for links (and several other things), but also black/bold/underline in some cases. The black/bold/underline seems too strong for the breadcrumb, which shouldn't really draw your eye from the main content.

adborden:
cc @thisisdano

line47:
I agree with you @adborden we don't want the breadcrumbs too strong. I like them as blue links and feel that reinforces that they are clickable and not just informational. Underlining those breadcrumb links could also strengthen that they are in fact clickable.

line47 (commented):
I think this looks good unless we have a class we can use to fit the content of the header that has already been established.

thisisdano:
1. I love this
2. Blue links w/ underlines seems like the way to go
3. wrt margin/panel, etc I'm going to take a look at the margin/padding on the page header today, as well as the margin/padding in marketplace and in panel rows in general, so hopefully by EOD we'll have away forward for a consistent solution for margins/padding.
4. I think we should merge it asap and go from there.

thisisdano (approved):
maybe make underline change, but otherwise let's try to get this live asap and work from there.

PR #262, opened 2017-01-03
Change dashboard width to 1008px (976px + 32px)

thisisdano:
This makes it the same as the sketch template — 976px active panel width + 16px (*2) outer margin.


line47 (approved):
Keep on truckin'... Looks good!

PR #263, opened 2017-01-03
PL doc updates for grid and reorganizing features

line47:
Separated the grid types into separate views within the PL. Moved count
status, count status grid and page header under features as they all
contain multiple components.


thisisdano (approved):
Very cool — I'm totally OK with this, but I wonder if we should split the flex grid classes out of the utility classes (as I'd had them originally) so we can use them in production code w/o extends.

This is not a discussion for this PR, tho — I'll bring it up separately.

thisisdano:
re: flex grid — I'm going to be thinking of this while I work on the marketplace panel (and on the page headers) today

PR #264, opened 2017-01-03
New logo image and button for top nav

thisisdano:
This PL updates the top nav to bring it in line with the Sketch file.

- Replaces cloud.gov logo image with dashboard-specific image
- brings Login / Log out button in line with global button style
- unrelated: reduces size of `.contents-primary` — it's now the same size as `.contents-secondary` yet I'm retaining it as a distinct style just in case we want/need some specific styling in the future

**Pair with https://github.com/18F/cg-dashboard/pull/907 in cg-dashboard**

- - -

screen shot 2017-01-03 at 10 00 51 am


line47 (approved):
This looks good here

PR #266, opened 2017-01-03
Moving more things around in the PL and updating statuses

line47:
Found a few more tweaks to make to the PL.


msecret:
Sorry I didn't get a full look on your last review, but I think there's some components/features I'd expect to be in `prototype` phase because they still contain the utility classes:

- Panel your orgs
- Panel overview org
- Panel, clean (just clean)
- Count status grid
- Card
- Trigger

Is that also what you see, I could be mistaken?

line47:
No problem, I was feeling that because we are currently using those items that they are at least `draft`. As our guidelines state for draft, "This pattern is a work in progress, may contain utility classes, and is not yet ready for production code. "

I'd say Trigger and Count Status Grid are still in `prototype` while the the other items should be in `draft`.

msecret:
Ah I think we should change the draft message to remove the `may contain utility classes`. I think it's important to make the distinction as utilities classes shouldn't ever show up on production, so if something has them, it shouldn't be a draft (as draft can be in production).

I think the confusion is that we took components that already existed, and moved them back to prototypes when we implemented them in the pattern library.

line47:
I agree with that. I'll make a few more updates to the PL.

msecret (approved):

PR #269, opened 2017-01-04
Updates to panel in the PL library and a few SCSS updates

line47:
Moved and renamed some panel related patterns under the panel section of the pattern library.
Updated some panel SCSS to streamline the panel style.

### DIFF
--- a/docs/src/components/features/panel/panel--org-expanded.html
+++ b/docs/src/components/features/panel/panel--org-expanded.html
@@ -1,8 +1,10 @@
-<article class="panel">
-  <header class="panel-header row-center-x">
-    <h1 class="panel-title col-fill">Your organizations</h1>
-    <div class="panel-secondary col-fit">
-      <span class="panel-secondary-description"></span>
+<section class="panel">
+  <header class="panel-header row-center-x mxn3 mtn3 mb3 p3 pb6 bb3 bb-cream">
+    <h1 class="panel-title col-fill">Organization overview</h1>
+    <div class="panel-secondary col-fit row-center-x">
+      {% render '@count-status--space'%}
+      {% render '@count-status--app'%}
+      {% render '@count-status--service'%}
     </div>
   </header>
   <div class="panel-content">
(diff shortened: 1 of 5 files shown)


docs/src/components/features/panel/panel--org-expanded.html
```diff
@@ -1,4 +1,4 @@
-<article class="panel">
+<section class="panel">
```
line47, line 1: Changing `<article>` to `<section>` as it has more semantic meaning for this situation.

src/css/components/panel.scss
```diff
@@ -97,18 +97,7 @@ h4 + .panel {
   align-items: center;
   display: flex;
   justify-content: space-between;
-  margin-top: $grid-4;
```
line47, line 100: Removing top margin on panel headers to bring styling closer to Sketch files.

src/css/components/panel.scss
```diff
@@ -97,18 +97,7 @@ h4 + .panel {
...
-  // TODO remove elements from this rule.
-  h1,
-  h2,
-  h3,
-  h4 {
-    font-size: $sans-s2;
-    font-weight: 700;
-  }
-
 }
```
line47, line 101: Removing H tags here as we are using the class `.panel-title` for defining.

docs/src/components/features/panel/panel.html
```diff
@@ -1,35 +1,37 @@
-<article class="panel">
-  <h1 class="panel-title">Panel title</h1>
-  <div class="panel-rows">
-    <section class="panel-group">
-      <header class="panel-header">
-        <h2>Group header</h2>
-        <span>Secondary information</span>
+<section class="panel">
+    <header class="panel-header row-center-x">
```
msecret, line 2: Can we remove `row-center-x`? Doing so could move this out of prototype. Here and elsewhere.

msecret (approved):

docs/src/components/features/panel/panel.html
line47: I'm working on another batch of updates that addresses this. (removes)
I have a bit more work to get through on it.

msecret: This LGTM then

PR #270, opened 2017-01-04
Style changes for header, breadcrumb,  and actions

thisisdano:
A few changes to improve appearance of the view header, overview panel, breadcrumbs, and actions...

- improves alignment and spacing in the header and breadcrumbs
- fixes a size error for action buttons
- fixes icon margin in the overview tree
- refines position/background of the loading indicator (but this still needs work in subsequent PRs)

**The breadcrumb will look a little funky until we see the changes in https://github.com/18F/cg-dashboard/issues/912**

- - -

screen shot 2017-01-04 at 10 20 57 am

### DIFF
--- a/src/css/components/breadcrumbs.scss
+++ b/src/css/components/breadcrumbs.scss
@@ -6,7 +6,8 @@
   @extend %fw400;
   @extend %sans-s5;
   @extend %lh-single;
-  @extend %mb2;
+  @extend %mb4;
+  @extend %mtn2;
 }
 
 .breadcrumbs-item {
--- a/src/css/components/breadcrumbs.scss
+++ b/src/css/components/breadcrumbs.scss
@@ -3,25 +3,31 @@
 @import '../core.scss';
 
 .breadcrumbs {
-  font-size: $sans-s3;
-  line-height: $lh-single;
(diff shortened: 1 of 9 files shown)


adborden:
If we're dropping the icon, we can get rid of the extra `-link_text` class, I'll push a change in a second.

src/css/components/breadcrumbs.scss
```diff
@@ -3,25 +3,31 @@
 @import '../core.scss';
 
 .breadcrumbs {
-  font-size: $sans-s3;
-  line-height: $lh-single;
-  margin: 0 $grid-3 $grid-3;
+  @extend %fw400;
```
msecret: Because we didn't come to a conclusion about whether we wanted to use the current utility classes, I don't really want to see code extended from them yet. This means here and elsewhere in this PR.

msecret (changes requested):

src/css/components/breadcrumbs.scss
thisisdano: awwww

thisisdano:
@msecret fixed yr issues

msecret (approved):

PR #271, opened 2017-01-04
Minor updates to panel between PL and prod code

line47:


src/css/components/grid.scss
```diff
@@ -1,7 +1,12 @@
 .grid {
-  @include outer-container;
```
line47, line 2: `outer-container` contained a max-width property that was not needed and causing issues in the PL.

docs/src/components/features/count-status/count-status--container.html
```diff
@@ -0,0 +1,5 @@
+<div class="count_status_container">
+  {% render '@count-status--space'%}
+  {% render '@count-status--app'%}
+  {% render '@count-status--service'%}
+</div>
```
line47, line 5: Adding container to count status

docs/src/components/features/panel/panel--org-expanded.html
```diff
@@ -1,10 +1,14 @@
...
+  <header class="panel-overview-header grid">
+    <div class="grid-width-8">
+      <h1 class="panel-title">Organization overview</h1>
+    </div>
+    <div class="grid-width-4">
+      <div class="count_status_container">
+        {% render '@count-status--space'%}
+        {% render '@count-status--app'%}
+        {% render '@count-status--service'%}
+      </div>
```
line47, line 11: Updating PL code to be more in line with what is in production

src/css/components/grid.scss
msecret: I think the correct thing here is to change the values [here](https://github.com/18F/cg-style/blob/master/src/css/vars.scss#L9) to the correct value. I think you might have to check that it doesn't break cg-site.

line47: In this case I don't want a max width at all.

msecret: I actually had to change this separately in my footer pr, https://github.com/18F/cg-style/pull/275. We can figure it out once they get merged together.

msecret (approved):

PR #273, opened 2017-01-04
Moved status marker to list start in PL

thisisdano:
Makes for a cleaner scan of PL status.

Oops, my panel clearfix is in here too. No matter — it has to happen anyway!

screen shot 2017-01-04 at 1 36 29 pm


line47 (approved):
This looks good to me

PR #274, opened 2017-01-04
Panel group style tweaks

thisisdano:
- added space between panel groups

Looks like this: 

screen shot 2017-01-04 at 2 29 03 pm


line47 (changes requested):
Please update the panel PL code to reflect this as well.
https://styleguide.fr.cloud.gov/components/detail/panel--default.html

thisisdano:
@line47 fixed, i think

line47 (approved):

PR #276, opened 2017-01-04
Update statuses for entity health

adborden:
Update the docs for Status. I spoke with @line47 and there is an underlying pattern we refer to as Status which is independent of Cloud Foundry application state. This pattern should be documented to be more generalized and then we can apply them to other contexts, like application health or activity logs.


line47:
Thanks for this @adborden, This is great!

line47 (approved):
This looks good

PR #277, opened 2017-01-05
Refined status grid prototype with sample prototype docs

thisisdano:
I fixed up my `count-status-grid` prototype to be a better example of delivering a prototype — with better logic and documentation.

@line47 thoughts? (esp on documentation...)

screen shot 2017-01-04 at 4 03 26 pm

### DIFF
--- a/docs/src/components/features/count-status-grid/README.md
+++ b/docs/src/components/features/count-status-grid/README.md
@@ -1,3 +1,33 @@
 # Count Status Grid
 
-Needs documentation...
+### Why is this pattern needed?
+A more descriptive count status would give a better quick read of the status of entities within a parent. It would more clearly show the number of entities in error.
+
+### Description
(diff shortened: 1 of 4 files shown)


docs/src/components/features/count-status-grid/README.md
```diff
@@ -1,3 +1,34 @@
...
+- **app-none** An app count with no apps
+- **app-ok** An app count with no apps in error
+- **app-error** An app count with some apps in error
+- **app-error-xxl** An app count with many apps in the parent
+- **service-none** A service count with no services
+- **service-ok** A service count with no services in error
+- **service-error** A service count with some services in error
+
+### Known issues
+The pattern begins to break down at ~20 entities, when the bars get very narrow.
```
line47, line 31: I'm wondering if there is a better way to display using a number in addition to the visual grid?

screen shot 2017-01-05 at 9 38 41 am

line47 (approved):
Looks good, see my comment about when the pattern breaks down.

PR #280, opened 2017-01-05
Add outer container neat back to grid

msecret:
Removing this breaks the grid because it removes the max width. I
checked this change on dashboard and site and seems to work.


line47 (approved):
Looks good

PR #281, opened 2017-01-05
PL updates for Icons, Icon UI and a few other minor updates

line47:
- Updated Icon UI with first pass at documentation
- Cleaned up Icons YML
- Minor naming updates for action and icon square components
- Fixed a few broken status references in the space treeview
- Added support for a dark background preview for light and inverse components.


docs/src/components/components/icon-ui/README.md
```diff
@@ -1,3 +1,24 @@
 # Icon Ui
 
-Needs documentation...
+The **icon ui** component is used to assist the user interface to help the user
```
msecret: `used to assist the user interface to help the user` sounds a little confusing. Maybe just `used to assist the user`?

msecret (commented):
My only other concern is I feel like the icon ui should be called fill icons based on how the class and the files are named, but that doesn't need to be fixed now.

docs/src/components/components/icon-ui/README.md
line47: Yeah that is definitely confusing.

msecret (approved):

PR #284, opened 2017-01-06
Panels, Columns, Rows

thisisdano:
**Work in Progress**
**Pair with https://github.com/18F/cg-dashboard/pull/929**


src/css/components/panel.scss
```diff
@@ -29,30 +28,32 @@ $panel-padding: $grid-3;
...
 .panel-title {
+  flex: 1 0 0;
   font-size: $sans-s7;
   font-weight: 400;
   line-height: $lh-single;
-  margin-bottom: $grid-3;
+  margin-bottom: $grid-5;
 }
 
 .panel-secondary {
```
msecret: What is panel secondary? What's it's purpose? I think defining that and naming it so it matches it's purpose might make it easier to use.

src/css/components/panel.scss
```diff
@@ -68,61 +69,44 @@ h4 + .panel {
...
 
 .panel-actions {
-  > * {
-    margin-right: 2rem;
+  margin-top: $grid-4;
 
-    &:last-child {
-      margin-right: 0;
-    }
+  .action {
```
msecret, line 80: Is there a way we can do this without nesting actions? There may not be and that's fine, just wanted to point out.

src/css/components/panel.scss
```diff
@@ -132,22 +116,16 @@ h4 + .panel {
...
-      margin-bottom: 0;
-      padding-bottom: 0;
-    }
+  &:last-child {
+    border-bottom: 1px solid $color-lightgray;
   }
 }
 
-.panel-row-boxed {
+.card {
```
msecret: Is there a reason this is here rather then in card component?

src/css/components/panel.scss
```diff
@@ -252,3 +230,14 @@ h4 + .panel {
   margin: (-$grid-3) (-$grid-3) $grid-3 (-$grid-3);
   padding: $grid-3 $grid-3 $grid-4;
 }
+
+// a couple random panel styles
+
+.route-link {
```
msecret: Similar comment to `card`, I'd prefer to keep only things related to `panel` in here so everything is easier to find and it keeps the current organization consistent. Is there a reason it should be in here though?

src/css/components/card.scss
```diff
@@ -7,7 +7,20 @@
...
+  > .panel-row-is_clickable {
+    display: flex;
+  }
+
+  &:last-child {
+    margin-bottom: 0;
+    padding-bottom: $grid-2;
+  }
+
+  .panel-row:last-child {
```
msecret: I like how you used `>` in the first rule above, can that be done here as well to keep more specific?

src/css/components/panel.scss
```diff
@@ -233,6 +227,18 @@ $panel-padding: $grid-3;
...
+  .panel-actions {
+    display: block;
+    text-align: left;
+
+    button:first-child {
+      margin-left: 0;
+      }
+    }
+}
+
```
line47: It was doing this:
screen shot 2017-01-06 at 2 45 41 pm

After this fix the buttons are aligned right and don't have as much space between the buttons.

msecret (commented):
Things are generally looking good, the thing I want to think about more is row, col and how you ahve to set start and end rows. Looking at the PR in dashboard, some stuff looks a lot more complicated to build now. Before you could just drop some panels, and now there's a lot more components and props you have to use to get the layout. I'm not saying that's bad, I just want to think about the component interface for developers.

I'm a fan of the card component though. I might try and find time to look at the col, row components and get a sense of how I'd use them and where things can be simplified.

src/css/components/panel.scss
line47: Correct me if I'm wrong @thisisdano but I believe that is for a panel subtitle

src/css/components/panel.scss
thisisdano: This panel is one of the next on my list!

src/css/components/panel.scss
thisisdano: This is for the information that goes at the top-right of the panel.

thisisdano: i.e. the logs message in the following panel:

screen shot 2017-01-06 at 12 29 01 pm

src/css/components/panel.scss
thisisdano: We could possibly do this with even margins and negative margin on the container (I like this), but this would just prevent the necessity of the `.panel-actions-right .action` nest. I don't _think_ you can apply the margin globally to action since the actions tend to appear in a wide-ish range of locations. This could be regularized, but it isn't regular yet.

src/css/components/panel.scss
thisisdano: oh huh — I thought I moved it.

thisisdano: ahh, moved it but never deleted it from `panel`

src/css/components/panel.scss
thisisdano: Agreed. I'd only done it because there weren't either `routes.scss` or `service-instances.scss` files, and I was hesitant to add them with only a single rule. This was kind of a kludge.

Should I just go ahead and add those two files?

src/css/components/panel.scss
thisisdano: Oops, nope. I tried this and the negative margin was great for the actions, but broke adjacent columns in a way I couldn't fix.

src/css/components/panel.scss
thisisdano: fixed

src/css/components/panel.scss
thisisdano: I'm happy to use a different name for the class if there are suggestions.

thisisdano:
@msecret 
You don't have to use start and end rows — those are alignment props, for setting the vertical alignment of the row (i.e. `flex-start`, `center`, and `flex-end`). Rows default to middle (I guess this should be `center`).

thisisdano:
I truly think this is a simpler system for devs — columns and rows are pretty straightforward for layouts, many things work well in the default state, and a few simple props open up more complex layout options. (Perhaps obviously) I personally find the new system to be simpler, cleaner, easier to read, and more predictable — tho ymmv.

I _am_ actively looking to improve this. :)

thisisdano:
Clarified those alignment props to be `valign` and used `top` `center` and `bottom` as values

docs/src/components/features/Tree View/space-tree-view.html
```diff
@@ -10,7 +10,7 @@ <h3 class="contents-secondary">
         </h3>
         </span>
         <span class="panel-column panel-column-less">
-        {% render '@status--running' %}
+        {% render '@status--ok' %}
```
adborden, line 13: Oops, didn't realize these were here. Thanks for fixing!

adborden (commented):
I'm a fan of having the row/col with clear rules. I'd caution about where panel is going, it seems like it's one component to rule them all. It handles some content related aspects like header/content/footer/actions, as well as presentation plain/boxed/lined. I could see us breaking down the panel pieces into smaller composable chunks.

thisisdano:
@adborden I'd agree that this leaves room to further component-ize special cases — `PanelRow` for instance instead of a `type`, and perhaps others. For now, this felt simpler _to me_, but I'm certainly open to more splitting and more components.

thisisdano:
OK — I've made a number of additions and improvements over the weekend:
- Simplified `Row`
  - Now only takes `valign` and `gutters` props
  - `panel-row` type now its own component: `PanelEntry` [documented in PL]
- new component `ContentsTreeColumn` for data columns in the contents tree [not in PL yet]
- Regularized contents tree components
  - `AppQuicklook` now `ContentsTreeApp` [documented in PL]
  - `SpaceQuicklook` now `ContentsTreeSpace` [documented in PL]
  - documented
- Changed block loading component
  - now the same as inline loading indicator
  - now can display loading text
  - made a regularized loading container (`.loading-container`) for loading in panels and cards. This prevents panel "jumping" on loading. [documented in PL]
- removed icon from Org in overview card
- removed the gray path from `ContentsTreeApp` [documented in PL]
- Changed `secondary` prop in `Card` (the top-right content) to `helperText` [documented in PL]
- Gave `main` a min-height to keep the footer at the bottom of the window

loading

Changing loading revealed that expand actions in the overview trigger loads on _all_ open cards.

We also get the disconcerting "double-load" on the Overview as well, with a quick load (perhaps with no status information) that leads to a second, more complete load in quick succession. The preliminary load feels unnecessary and confusing. (I see that this issue is already noted in https://github.com/18F/cg-dashboard/issues/869)

docs/src/components/components/card/README.md
```diff
@@ -0,0 +1,9 @@
+# Card
+
+A `Card` is a 100% width row with a 1px outside border and standard border radius. `Card`s typically live within a parent `Panel` and can be expandable. Use `Card`s to group sections of similar content.
+
+### Props
+none
+
+### Defaults
+none
```
thisisdano, line 9: Documenting new Card component

docs/src/components/components/card/card.html
```diff
@@ -1,9 +1,8 @@
 {% for org in organizations %}
-<div class="panel-row-boxed">
+<div class="card">
```
thisisdano, line 2: Using this pattern to show general Card rather than the Overview feature specifically.

docs/src/components/components/card/card.html
```diff
@@ -1,9 +1,8 @@
...
     <div class="panel-row-is_clickable">
         <div class="panel-column">
             <h2 class="card-title-primary">
-        {% render '@icon-square--org-medium' %}
```
thisisdano: No longer including icon in this view — it wasn't clear to users

docs/src/components/components/contents tree/contents-tree-app.config.yml
```diff
@@ -0,0 +1,15 @@
+title: ContentsTreeOrg (Contents tree / app contents)
```
thisisdano: Oops, this should be `ContentsTreeApp`

docs/src/components/components/contents tree/contents-tree-app.html
```diff
@@ -0,0 +1,33 @@
+{# ContentsTreeApp status="OK" #}
```
thisisdano: Oops. Can remove status from this note...

docs/src/components/components/loading/loading.html
```diff
@@ -1,8 +1,10 @@
-<div style="margin-top: 3.5rem"class="loading"
-    role="alertdialog"
-    ariaLive="assertive"
-    ariaBusy="true">
-  <img class="loading-indicator"
-      alt="Loading"
-      src="/img/loading.gif" />
+<div class="loading loading-container" role="alertdialog">
```
thisisdano: gotta add back the Aria info

thisisdano: done

docs/src/components/features/Tree View/space-tree-view.html
```diff
@@ -1,45 +0,0 @@
-<div class="panel-row">
```
thisisdano, line 1: replaced with `contents-tree`

docs/src/components/features/panel/panel.html
```diff
@@ -1,37 +1,40 @@
-<section class="panel">
-    <header class="panel-header row-center-x">
+<!-- `div` should be `section` -->
```
thisisdano: Brought this up to date w production code, but made a note where we should update production code semantics

docs/src/components/features/panel/panel.html
```diff
@@ -1,37 +1,40 @@
...
+      <div class="panel-entry">
+        <div>
+          <span class="col col-flex-1">
+            <span>panel entry</span>
+          </span>
+        </div>
+      </div>
     </div>
-</section>
+
```
thisisdano: brought `panel-actions` container up to date

docs/src/main.css
```diff
@@ -271,7 +271,6 @@ body {
   overflow: hidden;
   background-color: #d3d3d3;
   height: 16px;
-  width: 100%;
```
thisisdano, line 278: just changes for PL display

src/css/base/background.scss
```diff
@@ -8,5 +8,6 @@ body {
 
 main {
   background-color: transparent;
+  min-height: calc(100% - 22rem);
```
thisisdano, line 11: this keeps the footer at the bottom of the page

src/css/base/global.scss
```diff
@@ -7,7 +7,3 @@ body {
   font-size: 14px;
   margin: 0;
 }
-
```
thisisdano, line 10: dunno why we had this. it was wrong

src/css/components/all.scss
```diff
@@ -29,6 +29,7 @@
 @import 'sidenav.scss';
 @import 'search.scss';
 @import 'section.scss';
+@import 'service-instance.scss';
```
thisisdano, line 32: needed for a couple service instance panel styles

src/css/components/card.scss
```diff
@@ -7,7 +7,15 @@
   border: 1px solid $color-textblack;
   border-radius: $rounded;
   margin-bottom: $grid-2;
-  padding: 0;
+  padding: $grid-3 $grid-2;
```
thisisdano, line 10: regularized and passed flex on to the clickable div

src/css/components/contents.scss
```diff
@@ -21,3 +22,67 @@
 .contents-path {
   color: $color-mediumgray;
 }
+
```
thisisdano, line 25: All the new `contents-tree` styles

src/css/components/contents.scss
```diff
@@ -21,3 +22,67 @@
...
+.contents-tree-app,
+.contents-tree-space,
+.contents-tree-org {
+  width: 100%;
+}
+
+.contents-tree-space {
```
thisisdano, line 32: gets a little complicated to control nested border colors...

src/css/components/grid.scss
```diff
@@ -32,3 +32,77 @@
...
+.col,
+.row {
+  position: relative;
+}
+
+.row {
+  align-items: center;
+  flex-wrap: wrap;
+
+  // hack for react containing div requirement
```
thisisdano, line 45: these rules pass flex values on to unclassed child `div`s which are sometimes necessary in React templates

src/css/components/grid.scss
```diff
@@ -32,3 +32,77 @@
...
+  flex: 2 0 0;
+}
+
+.col-flex-none,
+.col-flex-0 {
+  flex: none;
+}
+
+$gutter: $grid-4;
+$gutter-offset: $grid-2; // $gutter / 2
```
thisisdano, line 86: I haven't had good luck getting math operators to work well in the sass — keep throwing errors — so I hardcoded

src/css/components/loading.scss
```diff
@@ -2,25 +2,66 @@
 $z-foreground: 200;
 
 .loading {
```
thisisdano, line 4: Absolutely positioned the block loader to get it to properly display full height of parent

src/css/components/loading.scss
```diff
@@ -2,25 +2,66 @@
...
   left: 0;
-  position: relative;
+  position: absolute;
+  right: 0;
   top: 0;
   width: 100%;
   z-index: $z-foreground;
 }
 
+.loading-container {
```
thisisdano, line 15: this is a container for the block loader if it is nested in something like a panel

src/css/components/loading.scss
```diff
@@ -2,25 +2,66 @@
...
+    display: none;
+  }
+}
+
 .loading-relative {
   top: auto;
 }
 
-$indicator-size: 3.5rem;
+.loading-information {
```
thisisdano, line 30: the new general container for the block loading indicator — sized so it can accommodate text

src/css/components/loading.scss
```diff
@@ -2,25 +2,66 @@
...
+    text-align: center;
+  }
+}
 
 .loading-indicator {
-  left: calc(50% - #{$indicator-size});
-  position: absolute;
-  top: calc(50% - #{$indicator-size});
-  width: $indicator-size;
+
```
thisisdano, line 55: a bigger loading indicator for the block loader

src/css/components/panel.scss
```diff
@@ -9,7 +9,6 @@ $panel-padding: $grid-3;
   @include clearfix;
   background-color: $color-white;
   margin-bottom: $grid-4;
-  // TODO make padding and margins consistent
```
thisisdano, line 12: TODID it

src/css/components/panel.scss
```diff
@@ -29,30 +28,32 @@ $panel-padding: $grid-3;
 
 }
 
-// TODO this is a hack, panel-row borders need to be reconciled, panel-row vs
```
thisisdano, line 32: TODID this

src/css/components/panel.scss
```diff
@@ -29,30 +28,32 @@ $panel-padding: $grid-3;
...
+  flex: 1 0 0;
   font-size: $sans-s7;
   font-weight: 400;
   line-height: $lh-single;
-  margin-bottom: $grid-3;
+  margin-bottom: $grid-5;
 }
 
-.panel-secondary {
+.panel-helpertext {
```
thisisdano, line 44: This is the new class for the helper text that appears in the top-right of a panel

src/css/components/panel.scss
```diff
@@ -68,61 +69,44 @@ h4 + .panel {
...
 
-  .action {
-    margin-left: 0;
-    margin-right: 0;
+  &.panel-actions-right {
+    float: right;
 
-    & + .action {
-      margin-left: .5rem;
+    .action {
```
thisisdano, line 87: made this a regular margin so no need for adjacency selector

src/css/components/panel.scss
```diff
@@ -68,61 +69,44 @@ h4 + .panel {
...
 .panel-row-header {
   font-size: $sans-s2;
   font-weight: 700;
   margin-bottom: $grid-1;
 }
 
-.panel-row {
-  $vertical-padding: $grid-105;
-  border-top: 1px solid $color-textblack;
+.panel-entry {
```
thisisdano, line 99: `panel-row` is now `panel-entry`

src/css/components/panel.scss
```diff
@@ -247,8 +177,22 @@ h4 + .panel {
...
 }
 
-.grid.panel-overview-header {
+.panel-overview-header {
   border-bottom: $grid-05 solid $color-cream;
   margin: (-$grid-3) (-$grid-3) $grid-3 (-$grid-3);
   padding: $grid-3 $grid-3 $grid-4;
 }
+
+// a couple random panel styles
```
thisisdano, line 186: gotta see if this is still necessary...

src/css/components/panel.scss
```diff
@@ -132,54 +116,19 @@ h4 + .panel {
     width: 100%;
   }
 
-  > .panel-row {
```
thisisdano, line 135: Amost all of these variations have been consolidated in to the `Row` component or into `PanelEntry` or `Card`

src/css/components/stat.scss
```diff
@@ -57,21 +57,38 @@ h5 + .stat {
...
+  font-weight: 400;
   white-space: nowrap;
 }
 
 .stat-single_box {
   $v-space: $grid-2;
-  background-color: $color-lightestgray;
-  margin: -$v-space 1.8rem;
-  padding: $v-space 0;
+  background-color: $color-cream;
```
thisisdano, line 72: All changes to make the Instances box display properly

src/css/components/stat.scss
```diff
@@ -57,21 +57,38 @@ h5 + .stat {
...
+  bottom: 0;
+  left: calc(50% - #{$grid-8});
+  margin: -$v-space 0;
+  padding: $v-space;
+  position: absolute;
+  right: calc(50% - #{$grid-8});
   text-align: center;
+  top: 0;
+
+  .subtext {
```
thisisdano, line 83: This may need an adjustment in the not-too-distant future...

thisisdano (commented):
My notes on these changes

PR #287, opened 2017-01-10
Minor style updates to forms and confirmations

line47:


src/css/components/actions.scss
```diff
@@ -6,7 +6,7 @@
 
 .action-link {
   color: $color-blue;
-  font-size: $sans-s3;
+  font-size: inherit;
```
line47, line 9: Set the font size of action links to inherit font size and not set explicitly. This allows for more flexibility.

src/css/components/confirm.scss
```diff
@@ -13,16 +13,15 @@
...
 
 .confirm-over {
   align-items: center;
   background-color: $color-lightestgray;
   border: 0;
   justify-content: space-between;
-  padding: 0.5rem;
-
+  padding: $grid-2;
 }
```
line47, line 25: Updating confirmation messages to use grid variables.

src/css/components/panel.scss
```diff
@@ -27,6 +27,19 @@ $panel-padding: $grid-3;
...
+    label {
+      color: $color-blue;
+    }
+
+    .route-form-preview {
+      font-size: inherit;
+      padding-left: 0;
+    }
+  }
+
```
line47, line 43: Setting background color to match the other form backgrounds. Also updating the label color to use a darker blue as the previous label color on the light gray background did not meet accessibility for color contrast.

Also updating the font size of the URL preview to inherit.

src/css/components/panel.scss
```diff
@@ -127,9 +140,9 @@ h4 + .panel {
   form {
     background-color: $color-background-light;
     flex-grow: 1;
-    margin: 1rem -1rem -1rem;
-    padding: 1rem;
-    width: 100%;
+    margin: 0;
+    padding: $grid-2;
+    display: block;
```
line47, line 146: Removing negative margin as ti was causing items to overlap in the layout. Utilizing grid variables.

src/css/components/panel.scss
```diff
@@ -189,9 +202,7 @@ h4 + .panel {
 }
 
 .panel-form-replace {
-  &.panel-form-replace {
-    margin: -1rem 0;
-  }
```
line47, line 194: removing styles that are not longer needed.

src/css/override_components.scss
```diff
@@ -135,15 +135,19 @@ button,
   }
 } /* stylelint-enable */
 
+.usa-button-disabled {
+  color: $color-textblack;
+}
+
```
line47, line 142: Setting the text color for disabled buttons to be accessible in terms of color contrast.

src/css/override_components.scss
```diff
@@ -135,15 +135,19 @@ button,
...
   outline: none;
 }
 
 .usa-label {
-  font-size: 0.7rem;
-  padding: 0.2rem 0.4rem;
+  font-size: $sans-s1;
+  padding: $grid-05 $grid-1;
   vertical-align: middle;
 }
```
line47, line 153: Updating `usa-label` to use the grid variables.

src/css/override_components.scss
thisisdano: This is the only thing I don't agree with. I don't believe that disabled buttons need to meet contrast guidelines. A disabled button should not look active, and contrast guidelines are there to assure that colors appear active.

This is not only personal belief, but a web standard. The W3C contrast guidelines explicitly exclude inactive UI:

> Incidental: Text or images of text **that are part of an inactive user interface component,** that are pure decoration, that are not visible to anyone, or that are part of a picture that contains significant other visual content, have no contrast requirement. (http://ux.stackexchange.com/questions/80881/do-disabled-buttons-still-need-to-be-contrast-compliant-for-accessibility)

I recommend that this button get `$color-white` text with `$color-mediumgray-trans` background. It's still a bit dark for my taste (I might even prefer $color-lightgray` bg), but gets the point across.

screen shot 2017-01-11 at 8 58 40 am

line47: Updated the disabled button styles.

src/css/components/panel.scss
msecret: Has this been tested on all dashboard pages?

src/css/components/panel.scss
msecret: Technically we shouldn't nest the route-form classes under panel, but we also don't have a place for the route form code, it currently resides in the dashboard CSS code. I guess I'm OK with this as long as we put a TODO about how this is temporary. Maybe like `TODO: remove once we abstract route form into own component`?

src/css/override_components.scss
```diff
@@ -135,15 +135,20 @@ button,
   }
 } /* stylelint-enable */
 
+.usa-button-disabled {
```
msecret, line 138: Where are we using this class?

src/css/components/panel.scss
line47: I'll add a comment

src/css/components/panel.scss
line47: It has been and no issues were found.

src/css/override_components.scss
line47: This class is used on the Restart app button for applications that are stopped.

screen shot 2017-01-11 at 1 52 03 pm

msecret: Hmmm, is it required for that button? or if using the `disabled` attribute on the button would be enough?

line47: taking the `usa-button-disabled` class off of the button the classes are `action action-primary` so the button looks active.

I could swap out the `usa-button-disabled` for `action-cautious` and that gets the same effect. 

As a side note, there a mix of styles that are a Frankenstein mix of `usa-` styles and cloud.gov styles.

msecret: Hmm, `action-cautious` isn't correct either. This might be a larger specificity issue we can't solve now. We can keep the `usa-disabled` class in.

msecret (approved):

adborden (approved):

PR #288, opened 2017-01-11
Adding examples of pages and update trigger

line47:
- Updated documentation for trigger component and moved it to feature as it includes multiple components. 
- Added some beginning work for Overview and Organization pages in the PL


docs/src/components/pages/org/org.html
```diff
@@ -0,0 +1,2 @@
+{% render '@page-header--with-icon' %}
```
line47: I'd like to know how to use this component but change the title and icon used in the header.
@adborden you might be able to help me solve that?

thisisdano:
I don't agree that the definition of a feature is that it uses multiple components. Typically a component is a component, even if it is built from other components. In this case, I consider `Trigger` to be a component.

My general thoughts are here: https://docs.google.com/spreadsheets/d/1jz_R-92lRCUcOqzHBB9D-5AQvO6U7fZHJ_AKMChIkXM/edit#gid=0

It's not 100% in sync with our current usage, but (from a design system standpoint) I consider a `feature` to be `a single unique presentation like search results` or like the Usage and Allocation panel, and a `component` to be `a single unique concept like a [single] search result`. This distinction is not hard and fast — there is some ambiguity — but if something has a single action or function, it's more like a component than a feature. `Panel` is a component but any specific panel is a feature.

line47:
That makes sense and I agree in this instance that `Trigger` should be a component.

docs/src/components/pages/org/org.html
adborden: :+1:

adborden: As a second argument to `render`, you can pass a JS object which will override properties in the component's `context`.

adborden (approved):

PR #292, opened 2017-01-30
Panel refactor start

msecret:
I recreated the simple parts of a panel. The panel has changed in
that it won't include the data part, the rows of information. It
will just include header, title actions and documentation.

I moved the feature of the panel, which will show a full example of the
panel with the rows to a separate name of panel full. I had to do this
because it seemed when you have folder/file of the same name in fractal
it gets them confused and you get the same result of both. I think they
should be separated to show how to use the panel in isolation from the
data rows which may be used alone.

I moved documentation into it's own component because it really acts as a hook
to style variable HTML a certain way. This allows a developer to wrap
any markup in documentation and it will all be styled consistently with
the panel.

This PR should keep the dashboard functioning without having to recode any
dashboard code. This means it can be safely merged any time.


src/css/override_vars.scss
```diff
@@ -10,6 +10,7 @@ $color-red:                     #be4900;
 
 $color-trueblack:               #000000;
 $color-textblack:               #333333;
+$color-textlighter:             #525252;
```
adborden: This seems a bit funny to me. Why are we adding a new color? I think I would prefer just to do one of the color [transform functions](http://sass-lang.com/documentation/Sass/Script/Functions.html#lighten-instance_method) to lighten one of the existing colors.

docs/src/components/components/panel/panel.html
```diff
@@ -0,0 +1,22 @@
...
+    </div>
+  </header>
+
+  <section class="panel-documentation panel-documentation-desc">
+    <p>Use this marketplace to create service instances for
+      spaces in this org. Then bind service instances to apps using the command line.
+      <a href="https://docs.cloud.gov/apps/managed-services/">Learn about using
+      service instances and marketplaces</a>.
+    </p>
+  </section>
```
adborden, line 16: I feel like in fractal terms, `panel-documentation` is a component and with some tweaking, could be used with the `{% render %}` syntax here.

adborden (approved):

docs/src/components/components/panel/panel.html
msecret: Whats the benefit of doing that?

adborden: Mostly just the DRY arguments: reduces the risk of changing in one place and forgetting to change it in another.

But also, in the view tab, fractal will also [link to the sub-component](https://styleguide.fr.cloud.gov/components/detail/trigger.html) which is helpful for docs.

screenshot from 2017-01-30 16-51-12

idk, this is a funny component where it is its own component by our definitions, but `panel-documentation` would never occur outside of a panel, would it?

msecret: It probably wouldn't

PR #293, opened 2017-01-30
Panel refactor documentation

msecret:
Started to remove some deprecated panel code from dep file and added
more rules for the panel actions to allow for different options.

This code doesn't cause any regressions as long as paired with
https://github.com/18F/cg-dashboard/pull/990.

### DIFF
--- a/src/css/components/panel.scss
+++ b/src/css/components/panel.scss
@@ -40,6 +40,18 @@ $panel-padding: $grid-3;
   flex-direction: row;
   justify-content: flex-start;
   margin: 0;
+
+  p, span {
+    font-size: 0.9rem;
+  }
+}
+
+* + .panel-actions {
+  margin-top: $grid-1;
+}
+
+.panel-actions + * {
+  margin-top: $grid-2;
 }
 
 .panel-actions-right {
--- a/src/css/components/panel-dep.scss
+++ b/src/css/components/panel-dep.scss
@@ -21,32 +21,6 @@ $panel-padding: $grid-3;
   }
 }
 
-.panel-actions {
-  display: block;
-
-  > * {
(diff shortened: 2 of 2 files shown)


src/css/components/panel.scss
```diff
@@ -40,6 +40,18 @@ $panel-padding: $grid-3;
...
+    font-size: 0.9rem;
+  }
+}
+
+* + .panel-actions {
+  margin-top: $grid-1;
+}
+
+.panel-actions + * {
+  margin-top: $grid-2;
```
adborden, line 54: Isn't the structure of panels fairly known that we don't really need these `*` rules? I mean, we could say something like `.panel-documentation + .panel-actions` rules instead?

msecret: OK we can keep it more specific like that.

msecret: Actually now that I've worked on the panel rows, it's not that clear. It could be rows, actions, documentation, etc.

adborden: Yeah, I figured there would be several, but still known and you would end up with something like
```scss
.panel-actions {
  & + .panel-rows,
  & + .panel-actions,
  & + .panel-documentation {
  /* ... */
  }
}
```

msecret: But it's not `panel-rows` anymore, and panels allow for any type of content besides the documentation and actions

adborden (approved):

src/css/components/panel.scss
adborden: I'm not a fan of `*` rules, but given that this is simply adding margin between elements it seems sane enough. I've certainly seen similar elsewhere.

PR #295, opened 2017-02-06
Refactor usage and limits

msecret:
By removing panel code and making the stat component much more
specific. Rather then making the stat component able to adapt
to many different sizes, it sets all the sizes of things. This
makes it much easier to build and I'm not sure the adaptability
will be required for this component anyway.

### DIFF
--- a/src/css/components/page_header.scss
+++ b/src/css/components/page_header.scss
@@ -16,6 +16,7 @@
 
 .page-header-title {
   flex: 1;
+  min-width: 26rem;
 
   .icon-large {
     top: -3px;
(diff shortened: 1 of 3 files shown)


src/css/components/page_header.scss
```diff
@@ -26,4 +27,8 @@
   display: flex;
   flex-direction: row;
   margin-right: -$grid-1;
+
+  .error-inline {
+    width: auto;
+  }
```
adborden, line 33: I feel like this is the common case, we want `.error-inline` to be `width: auto` in most places. I think the one exception is in form elements, but I'm not sure. Can we un-nest this rule?

adborden (approved):

PR #300, opened 2017-02-08
Fix color contrast issues in text

msecret:
The success button green is too light and the gray for the paths
in the quicklooks is too light of a gray.

Dano mentioned he wanted a different solution for the green but I
think we should fix accessibility issues as soon as possible.

Refs https://github.com/18F/cg-dashboard/issues/921

### DIFF
--- a/src/css/components/contents.scss
+++ b/src/css/components/contents.scss
@@ -19,5 +19,5 @@
 }
 
 .contents-path {
-  color: $color-mediumgray;
+  color: $color-darkgray;
 }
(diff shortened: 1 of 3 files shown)


adborden:
I'm okay with the grey space text change, but the green color looks too dark. Especially for the icon and the status-pill. I think it's okay to go dark green for contrast regarding text, but I don't think we should make the change across the board.

Generally I don't think any of the green text is critical for accessibility so I would rather hold off on the green change for now and wait until we have a more complete color palette solution.

msecret:
Even though the green status text isn't that important, I'm still not sure I want to leave it inaccessible. If somebody has trouble reading it, they might not know it's not important. I separate the green text color from the rest of the green.

adborden (approved):

PR #302, opened 2017-02-11
UX fixes for activity log

adborden:
- Remove flex styles, now handled by elastic_line component
- Remove status pill on all events except for error
- Move status pill to be anchored on  the line rather than log item
- Manually calculate height/top for vertical alignment
- Remove extra background color

One thing that sticks out is that elastic_line is embedded within the activity_log component. This is more an artifact of React because ElasticLine is it's own component. If we weren't using React, I could imagine activity_log items `@extend`ing elastic_line instead of embedding it.

Another callout, the vertical align of the status pill used to be done by the `flex` layout, but now it's done using height/offset calculations. Let me know if there is a better way.

### DIFF
--- a/src/css/components/activity_log.scss
+++ b/src/css/components/activity_log.scss
@@ -11,34 +11,30 @@
   width: 100%;
 }
 
+.activity_log-item_line {
+  cursor: pointer;
+  font-family: $font-mono;
+  font-size: $sans-s3;
+  padding: $grid-1 0 $grid-1 $grid-2;
+  position: relative; // For status-pill anchoring
+}
+
 .activity_log-item_text {
-  position: relative;
   word-wrap: break-word;
 }
 
 .activity_log-item_timestamp {
-  cursor: pointer;
   text-align: right;
-}
 
-.activity_log-item_text,
-.activity_log-item_timestamp {
-  flex-grow: 1;
-  font-family: $font-mono;
-  font-size: $sans-s3;
-  max-width: initial;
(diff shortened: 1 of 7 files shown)


src/css/components/activity_log.scss
```diff
@@ -48,50 +44,35 @@
...
-    @extend .status-pill !optional;
-    height: 75%;
-  }
-
-  .activity_log-item_text {
-    font-family: $font-sans;
-    font-weight: 600;
+  .activity_log-item_line {
+    .activity_log-item_text {
+      font-family: $font-sans;
```
msecret: Why is so much nesting needed? If it's hard to override, can we fix what's so high in specificity?

docs/src/components/components/activity-log/activity-log.html
```diff
@@ -0,0 +1,40 @@
+<ul class="activity_log">
```
msecret: I think the activity log should belong in `features` because in order to render correctly, it needs to contain other components like `elastic_line` but I'm open to it belonging here too.

src/css/components/activity_log.scss
adborden: It's from `.panel span`. I didn't want to touch it until after the panel refactor but I did add a comment about it.

docs/src/components/components/activity-log/activity-log.html
adborden: :+1:

msecret (approved):

PR #308, opened 2017-02-21
Global notification and button refactoring

msecret:
This creates a new prototype notification feature. This is meant to provide feedback about errors and other things to the user on a global level. It's required for the global error handling in the dashboard. It has different parts to it, wrap, message, actions, and dismiss. It also has modifiers; warning, info, error, finish, that link to the existing states in the design system.

The button code was refactored to make it easier to add the action buttons to the notification panel. No code needs to be changed for this to work, but eventually we should remove `usa-button` code in both cg-site and cg-dashboard. This should be done in a separate PR and go into a large release.

### DIFF
--- a/docs/src/components/_preview-bg-dark.html
+++ b/docs/src/components/_preview-bg-dark.html
@@ -2,7 +2,7 @@
 <html>
 <head>
     <meta charset="utf-8">
-    <link media="all" rel="stylesheet" href="{{ '/css/cloudgov-style-min.css' | path }}">
+    <link media="all" rel="stylesheet" href="{{ '/css/cloudgov-style.css' | path }}">
     <link media="all" rel="stylesheet" href="{{  '/css/main.css' | path }}">
     <title>Cloud.gov Pattern Library</title>
 </head>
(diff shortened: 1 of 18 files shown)


docs/src/components/_preview-bg-dark.html
```diff
@@ -2,7 +2,7 @@
 <html>
 <head>
     <meta charset="utf-8">
-    <link media="all" rel="stylesheet" href="{{ '/css/cloudgov-style-min.css' | path }}">
+    <link media="all" rel="stylesheet" href="{{ '/css/cloudgov-style.css' | path }}">
```
adborden, line 5: Hmm, does this simplify the build at all? I think we can drop the `compile-css` step from https://github.com/18F/cg-style/pull/308/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R16

adborden (approved):

docs/src/components/_preview-bg-dark.html
msecret: Removed that, good idea.

PR #312, opened 2017-03-17
Styles for Info components

adborden:
Some of the default styles have rules like `p + h5` which break with the info component because it contains things other than `p`. Using class-based rules avoid that.


adborden:
@msecret I suspect this isn't needed anymore?

PR #313, opened 2017-03-21
Expressive empty states

msecret:
Adds a new component to deal with specific empty state UI case. Modifies existing panel documentation to deal better with empty states.

image


rememberlenny (approved):
Looks good!

PR #315, opened 2017-06-05
Add some scroll power to code blocks

rogeruiz:
There are a lot of code-blocks in the Operations parts of the documentation that are illegible with word-wrapping. We need to be able to scroll our code-blocks in style. :sunglasses:


rogeruiz:
Looks like it made it into `master` previously (18F/cg-style#98.), but was never published? either way 👍 let's get this in please.

msecret (approved):

PR #317, opened 2017-06-20
change text-transform for status to none

suprenant:
Intended to fix this issue, where we have confirmation dialogs with sentences with every word capitalized, which is awkward to read:
https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-5851

This is my first CSS change on the project, so please let me know if this fix isn't right.

### DIFF
--- a/src/css/components/status.scss
+++ b/src/css/components/status.scss
@@ -4,7 +4,7 @@
 @import '../core.scss';
 
 .status {
-  text-transform: capitalize;
+  text-transform: none;
 }
 
 .status-started,
--- a/src/css/components/status.scss
+++ b/src/css/components/status.scss
@@ -3,10 +3,6 @@
 @import '../override_vars.scss';
 @import '../core.scss';
 
-.status {
-  text-transform: none;
-}
-
 .status-started,
 .status-running,
 .status-ok {


fureigh (changes requested):
@suprenant Thanks for the PR! Nice work tracking down the source of the unwanted capitalization.

This raises an important question about the level on which we want to make the change. (The joys of CSS!) The usual question is: what's the smallest possible change to achieve the desired effect without adverse unintended consequences?

Specifically, in this case (pun not intended): if you change the styling of `.status` here, you'll change it for any and all uses of that class in any and all projects that use the cloud.gov style guide. Is that what you intended? If so, what else will that change affect?

The `status` class is applied to this particular element by https://github.com/18F/cg-dashboard/blob/master/static_src/components/create_service_instance.jsx#L110. The `.status` class is currently used one other place in that project:
https://github.com/18F/cg-dashboard/blob/master/static_src/components/app_quicklook.jsx#L47

Were @msecret in this week, I'd ask him what his intent was, as `git blame` identifies him as the original author of this line. As he's not, I recommend searching the git log for the commit in which this was added, in hopes of understanding what the reason was for writing this style rule.

If that reveals that this style rule should go, let's entirely delete lines 6 through 9 rather than changing `capitalize` to `none`. (Doublecheck, but I suspect that as-is it isn't overriding anything.) You'll want to make that change with `npm run watch` running per https://github.com/18F/cg-style/blob/master/CONTRIBUTING.md#L28, though. Ping me and we can pair on that.

If instead 1) we want to keep this capitalization option in https://github.com/18F/cg-style but 2) we still don't want it applied to the text in https://github.com/18F/cg-dashboard/blob/master/static_src/components/create_service_instance.jsx#110, let's close this PR in favor of one in `cg-dashboard` and look for a more suitable style to apply instead.

fureigh:
@suprenant ^ The lengthy review is because this is a proposed CSS change to a style library, which has different implications than a simple copy edit on a webpage. Slack me if you'd rather talk through it in realtime. Thanks for your willingness to dive into this!

suprenant:
@fureigh thanks for the thorough and deeply educational review! I put time for us on Marco's calendar once he's back so we can ask his input and then take next steps here.

msecret:
@fureigh the only reason I could imagine doing uppercase transform is for design reasons. I unfortunately can't see the issue in Favro though (maybe it doesn't want me to login from Portugal?) so I can't really understand the issue.

I would say the right solution is to remove that role completely from cg-style. We want to minimize overrides as much as possible in cg-dashboard and keep style consistency in cg-style.

suprenant:
screen shot 2017-06-19 at 4 25 38 pm

fureigh:
s/role/rule, got it!

suprenant:
@fureigh @msecret and I talked through this – Fureigh is going to remove the .status class locally, Marco will review, and we should be done with this from there.

fureigh (approved):

fureigh:
Ready for your 👀, @msecret.

msecret (approved):

PR #325, opened 2017-08-15
adjusted nav sizes, added ext link icon, adjusted search icon

femmebot:
Style updates related to https://github.com/18F/cg-site/pull/1014#issuecomment-322538716
- small adjustments to nav links (so it doesn't feel so crowded)
- added external link icon so users know what they're clicking on an external link (and will lose cloud.gov nav/context) 
- adjusted the search icon weight so it matches the font weight of the nav links

cc @el-mapache


src/css/components/nav.scss
```diff
@@ -37,16 +37,24 @@
...
+
   & + .nav-link {
     padding-left: 0;
 
     @include media($medium-screen) { padding-left: $grid-4; }
   }
 
   @include media($medium-screen) {
-    font-size: inherit;
+    // font-size: inherit;
```
el-mapache: I think we can safely delete this line

jcscottiii: you have a comment here. should remove it

src/css/components/nav.scss
```diff
@@ -37,16 +37,24 @@
     &:hover {
       color: $color-primary;
     }
+    &[href *= ".io"] {
```
el-mapache: Do you expect that this icon will only be used in the context of an element with a `nav-link` class? If not, we might try putting this icon somewhere more generic.

el-mapache (commented):
Thanks for adding these! I don't see the icon added in the diff, can you double check and make sure it was committed?

I just left a couple of teeny comments!

src/css/components/nav.scss
```diff
@@ -37,16 +37,24 @@
...
+      padding-right: 22px;
```
jcscottiii: typically we try to use rem where possible

src/css/components/nav.scss
femmebot: Yeah, I was thinking of genericizing it so that all off-domain links are appended with the external-link-icon but thought that might be way outside the scope? Maybe something we can refactor/address separately?

el-mapache: Sure, we can address it in a refactor once we figure out what the use cases will look like in the future.

el-mapache: Looks like our css linter doesn't like the use of qualified selectors here (the `href *= ".io"`) so I think we should just move it to a separate class after all.

jcscottiii:
@femmebot the `/img` folder shouldn't be used. instead you want to use `src/img`

src/css/components/nav.scss
femmebot: It's gone!

src/css/components/nav.scss
femmebot: Okay, I changed it to rem. (and it _does_ look like 16 px = 1 rem)

src/css/components/nav.scss
femmebot: okedokee. i just updated it and changed it to a class `.external-link`

femmebot (commented):
@jcscottiii yup, `src/img/` is where I have the images. looks like .gitignore might be ignoring both `src/img/` and `/img/`?

src/css/components/nav.scss
```diff
@@ -37,16 +37,23 @@
...
     }
+    &.external-link {
+      padding-right: 1.375rem;
+      background: url('#{$static-img-path}/external-link-grey.png') right center no-repeat;
+      background: url('#{$static-img-path}/external-link-grey.svg') right center no-repeat;
+      background-size: 15px 15px;
+    }
   }
 
+
```
el-mapache: It looks like the linter also wants this blank line deleted!

src/css/components/nav.scss
```diff
@@ -37,16 +37,23 @@
     &:hover {
       color: $color-primary;
     }
+    &.external-link {
```
el-mapache: Since this is nested within an `a`, and uses the `&`, this also throws a qualifying type error in the linter, since the css resolves to `a.external-link`. I would put the `&.external-link` declaration directly inside the `nav-link` body.

src/css/components/nav.scss
femmebot: wut???

el-mapache: Yep, our linter is set to only allow 1 blank line between declarations. If you click the `details` link on the circle ci integration below, you can see exactly what lines are considered problems.

I'd recommend running `npm run lint` locally after you make changes, then you can see the linting errors as they happen!

src/css/components/nav.scss
```diff
@@ -27,26 +27,31 @@
...
+  font-weight: 500;
   padding-bottom: .2rem;
   padding-left: 0;
   padding-top: .2rem;
+  .external-link {
+    padding-right: 1.375rem;
+    background: url('#{$static-img-path}/external-link-gray.png') right center no-repeat;
+    background: url('#{$static-img-path}/external-link-gray.svg') right center no-repeat;
+    background-size: 15px 15px;
+    }
```
el-mapache: This should be indented the same as `.external-link`. The linter is VERY picky 😄

jcscottiii:
sorry for the confusion
this file is in `cg-style/img` and should be moved to `cg-style/src/img`
image
same thing with these
image

@femmebot

src/css/components/nav.scss
femmebot: eep. ok, i just ran the linter and didn't find any more errors. thanks so much for your patience!

src/css/override_components.scss
```diff
@@ -185,8 +185,8 @@ input:focus {
 
   [type='submit'] {
     background-color: #ffffff;
-    background-image: url('#{$static-img-path}/search.png');
```
jcscottiii, line 188: i don't think we need `search.png` or `search.svg` anymore now right? i think we can delete them now.

femmebot: Gone!

jcscottiii (approved):

PR #328, opened 2017-09-06
updated style for Homepage layout with promo section

femmebot:
See https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-8323


src/css/override_vars.scss
```diff
@@ -3,7 +3,7 @@
 // named colors
 $color-blue:                    #056dd4;
 $color-brightmint:              #00df99;
-$color-cream:                   #f5f5f3;
+$color-cream:                   #f5f5f3; // formerly #f8f5f0
```
jcscottiii: no need for the comment

src/css/components/section.scss
```diff
@@ -120,10 +120,14 @@ $chevron-size: 4.865rem;
...
-    text-decoration: underline;
+    // color: $color-red;
+    // text-decoration: underline;
+    color: $color-blue;
+    text-decoration: none;
+    border-bottom: 1px solid $color-blue;
   }
-
+  // these seem to just override the definitions immediately above it
+  // why keep both?
```
jcscottiii: this is is so they don't have to copy the text-decoration and border-bottom part to both sets.
the second set is just to override the color.

jcscottiii:
Also, could you address the build failures?

jcscottiii (changes requested):
Address the comments and the build failures

jcscottiii:
Also, is there a reason you change the perms on the files?
image

I don't think you need execute permissions on these files.

jcscottiii:
@femmebot take a look at these lines: 
image

Also for the permission thing, do `chmod 644 filename`

jcscottiii (approved):

PR #335, opened 2018-01-19
Let's go to node 8 and fix dependencies

pburkholder:
node 8 avails us of better dependency management with npm 5.5, so updating package-lock.json is a lot easier.

This fixes the vulnerable dependency warning we get on GitHub.

Once this is done we can cut a new release and then point cg-site to that release.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,7 +1,7 @@
 
 machine:
   node:
-    version: 6
+    version: 8
   environment:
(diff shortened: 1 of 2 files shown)


pburkholder:
@msecret I've tagged you on this review to chime in if there are Node 8 gotchas that we should be aware of. Thanks! P.

femmebot (approved):
YOLO!

PR #339, opened 2018-12-13
Release version 2.11.0

rogeruiz:
This release includes the latest USWDS upgrade until 2.0.0 is released. We probably don't want to keep going after that since it's going to introduce breaking changes. This work fixes the search bar and flag issues. From now on, we should be getting minor and patch fixes from the upstream USWDS repo.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+package-lock.json binary
--- a/circle.yml
+++ b/circle.yml
@@ -25,9 +25,3 @@ deployment:
       - npm run build-library
       - scripts/deploy-library.sh cg-style
 
-  npm:
-    tag: /[0-9]+(\.[0-9]+)*/
-    owner: 18F
-    commands:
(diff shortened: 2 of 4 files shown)


rogeruiz:
oh ffs, @msecret was removed from the npm package and so the CI is failing because it used his credentials to login. I'll fix this in the morning. :pour-one-out: for Marco. I wonder if we'll be able to just remove the npm login work entirely. At first glance it looks like it's publishing via CI. So we don't need it. 💯 🎉

rogeruiz:
also, @msecret not to worry. I've gone ahead and deleted your credentials from our CI build environments so no one can access your creds via a `re-run with SSH` command. I would rotate your password though just in case. 😉

rogeruiz:
This package has been published prior to this being merged into master. https://www.npmjs.com/package/cloudgov-style/v/2.11.0

mogul (approved):

PR #341, opened 2019-01-16
Config for CircleCI 2.0

adborden:


siennathesane (commented):
I think this is okay, I don't see a reason not to use CircleCI, could you provide some commentary on the reason?

adborden:
cg-style was using the old 1.0 config format which has been sunset. This just brings it up to the 2.0 config version.

siennathesane (commented):
I'm fine with this but I would like to see it in action.

siennathesane (approved):
LGTM

PR #354, opened 2019-09-09
remove package-lock.json from repository, as this code is consumed as a package

afeld:


siennathesane:
@afeld I'm not a JS person, but isn't `package-lock.json` also used for this code's internal dependency packaging? I don't think we can remove it.

afeld:
No, it would only be used when running `npm install` from within this repository specifically.

> One key detail about package-lock.json is that it cannot be published, and it will be ignored if found in any place other than the toplevel package.

https://docs.npmjs.com/files/package-lock.json

[Ongoing thread in Slack](https://gsa-tts.slack.com/archives/C032KSPPQ/p1568039930001600)

siennathesane:
Is it breaking something downstream?

